PR #44, opened 2015-06-15
Change instance type from m1.small to t2.small

menicosia:
Suggest changing default instance type in the MicroBosh AWS manifest from m1.small, which does not support HVM to t2.small, which does, and are less expensive. (HVM stemcells are smaller to dl/ul.)

### DIFF
--- a/deploy-microbosh-to-aws.html.md.erb
+++ b/deploy-microbosh-to-aws.html.md.erb
@@ -37,7 +37,7 @@ network:
 resources:
   persistent_disk: 20000
   cloud_properties:
-    instance_type: m1.small
+    instance_type: t2.small
     availability_zone: AVAILABILITY-ZONE # Replace with your Availability Zone
 
 cloud:


amitkgupta:
The t2.small are less expensive, but don't come with any storage, you need to purchase an EBS volume.  There are different kinds of EBS volumes, and you may not need the full 150GB that comes with the m1.small, but you could easily end up paying more.  The t2's also have Burstable CPU performance, which should be okay, but it depends on the usage.

If stemcell upload/download times are the concern, why not use "light" stemcells?

menicosia:
Hi @amitkgupta,

You are right. I was accidentally comparing a light Xen-HVM to a plain Xen stemcell.

I guess that definitely makes this PR less compelling, except that Amazon tends to be heading in the direction of HVM non-local storage VMs, and the m1.smalls are considered 'legacy.'

I did check on Amazon's pricing calculator - a t2.small runs about $19/mo while an m1.small runs about $32/month. A regular EBS volume of 160GB runs about $8/month, so a meager savings of about $5/month.

Shrug, sysadmin would get an extra donut each month. Feel free to reject this PR. :)

PS - If reserved instances, the difference in price is only .003/hr, making the m1.small less expensive than the t2.small when you add in EBS!

kimahoff:
@cppforlife The docs team will leave this one for you to merge or reject-- there's no editing or anything that would call for docs to get involved.

cppforlife:
Changed to m3.xlarge.

PR #45, opened 2015-07-10
update releases

ronakbanka:


cppforlife:
Updated all releases. Thanks!

PR #46, opened 2015-07-15
Putting properties in the spec file is mandatory

voelzmo:
Properties that are not in the spec file are not available in the template expansion, right?

### DIFF
--- a/create-release.html.md.erb
+++ b/create-release.html.md.erb
@@ -716,12 +716,6 @@ properties:
      default: 80
 ~~~
 
-Note that the definition of properties and their defaults in the job's spec file is optional:
-Bosh will still expand properties in templates and update relevant jobs upon modifications of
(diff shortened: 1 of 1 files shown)


gberche-orange:
@voelzmo @cppforlife 

From the thread at https://groups.google.com/a/cloudfoundry.org/forum/#!msg/bosh-users/26pogGElU5s/KDFPkfibdUQJ I experienced that properties not declared in the Spec file were indeed available for template expansion, hence pr https://github.com/cloudfoundry/docs-bosh/pull/40

Has bosh changed this behavior lately, or did I misunderstood something ?

voelzmo:
Is that just working because the templates are not using the `p('some_property')` notation? I noticed that the cf-services-contrib-release seems to be accessing the `properties` object directly. @cppforlife could you maybe clarify this?

PR #52, opened 2015-09-23
Added 'patch' to list of Centos dependencies

vestel:
I have several cases, bosh-init for AWS fails, when I use fresh new Amazon Centos 7 VM.
The failure appears on bosh_aws_cpi compilation, because system could not build native libxml extentions for nokogiri gem.


cppforlife:
Updated. Thanks.

PR #54, opened 2015-09-28
Adding possible errors and fixes for OS X.

bridgetkromhout:
Hi, folks. I was working through http://mariash.github.io/learn-bosh/#install_bosh_cli on a box I'd recently upgraded to Yosemite and hadn't yet used Xcode on. The error message was pretty misleading (and unjustly maligned nokogiri, which usually deserves it). Figured that even though Xcode is mentioned elsewhere, it might be good having this explanation be more clear on the page you end up at if you have a failure of `gem install bosh_cli`.

This fix covers: 
    - Missing Xcode cli tools
    - Lack of signed Xcode license
The error message for these is misleading, so it's probably worth showing the output explicitly.

Oh, and I work for Pivotal, which I imagine covers the CLA.

### DIFF
--- a/bosh-cli.html.md.erb
+++ b/bosh-cli.html.md.erb
@@ -87,6 +87,59 @@ Make sure following packages are installed:
 $ sudo yum install gcc ruby ruby-devel mysql-devel postgresql-devel postgresql-libs sqlite-devel libxslt-devel libxml2-devel yajl-ruby
 </pre>
 
+### Prerequisites on Mac OS X
+
+You may see an error like this:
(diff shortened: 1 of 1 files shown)


cppforlife:
Cool. We'll pull it in. 

I would recommend making Pivotal org association visible in github account to avoid getting CLA comments.

bridgetkromhout:
> Cool. We'll pull it in.

Thanks!

> I would recommend making Pivotal org association visible in github account to avoid getting CLA comments.

How do I get invited to the appropriate org? :)

animatedmax:
Thank you, bridgetkromhout !

PR #56, opened 2015-10-12
Remove --skip-if-exists from release examples

benmoss:
This is a deprecated no-op now, so we shouldn't make people think it does anything.


animatedmax:
Thanks, benmoss

PR #57, opened 2015-10-16
Generate private key and cert request in two steps

holgerkoser:
On Ubunutu with OpenSSL 1.0.1f using `openssl req` with option `-newkey rsa:2048`
the generated private key has PKCS#8 format -----BEGIN PRIVATE KEY-----
which is not accepted by UAA.
With `openssl genrsa` the generated private key has PKCS#1 format
-----BEGIN RSA PRIVATE KEY----- which is accepted by UAA.

### DIFF
--- a/director-certs.html.md.erb
+++ b/director-certs.html.md.erb
@@ -36,10 +36,13 @@ extensions = san
 subjectAltName = IP:${ip}
 EOL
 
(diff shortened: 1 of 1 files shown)


animatedmax:
Thank you, holgerkoser !

PR #63, opened 2015-10-29
details on bosh-agent interactions with the CPI

gberche-orange:
adds:
- config formats documentation
- interactions overview diagrams

This reuses material prepared for http://sched.co/4Nzh
Let me know if cloudstack specific example are an issue or if you see incorrect elements that should be fixed (in this PR and in [our slides](https://drive.google.com/file/d/0B_RQz82RzSUnUHBEdjNJLXc2S0E/view?usp=sharing))

Thanks in advance for your feedback,

Guillaume.

/CC @poblin-orange


gberche-orange:
Thanks @cppforlife for the prompt merge and the formatting cleanup in https://github.com/cloudfoundry/docs-bosh/commit/ed28a2d9495678a9530cff9af467359ea895bbd3 !

Can you please remind me if there is a public CI which pushes this repo changes onto http://bosh.io/docs as I don't yet see it updated, and was eager to check resulting formatting (I did not yet have time to setup bookbinder to generate docs-bosh html output) ?. 

BTW in https://github.com/pivotal-cf/bookbinder/issues/105#issuecomment-150296491 @kimahoff thought http://bosh.io isn't using bookbinder (and hence would'nt leverage text-based formatting added to it), is that also true for http://bosh.io/docs ?

PR #68, opened 2015-11-26
Fix typos and copy/paste errors

voelzmo:


cppforlife:
i didnt include features on purpose since i want to deprecate that section. did you need it for something?

voelzmo:
That's what I suspected, wanted to run it by you anyways :) I'll remove that part from the PR then, fair enough.

Don't need the features in that call, I suspect you move that information to another call or property then? Checking if snapshotting is enabled is a good thing to have for me.

cppforlife:
> Checking if snapshotting is enabled is a good thing to have for me.

What's use case for checking that?

voelzmo:
>> Checking if snapshotting is enabled is a good thing to have for me.

> What's use case for checking that?

I just want to make sure the director I'm looking at has snapshots enabled or not. This seems to be one of the things people are unsure about when talking about "their" director. Resurrection is imho another candidate for exposing if enabled or not. This might be a very selective view, but I find myself checking those two things quite a bit.

PR #72, opened 2015-12-15
Fix a link to Resource Pool in Terminology

wlindner:


wlindner:
I made my organization membership to cloudfoundry public, so I should no longer need to sign the CLA.

dpb587-pivotal:
I think `resource-pools` was the correct form? It's linked from other lines in the file there, too?

cppforlife:
@wlindner plural is correct here. is there a link that points to #resource-pool somewhere?

wlindner:
@dpb587-pivotal @cppforlife my mistake, I made the change in the wrong place. Here's the PR for fixing the link: #74

PR #91, opened 2016-02-15
Attempt to make the examples match their description

Fydon:
The description before the examples state "example of a VM with 1 CPU", but then the cloud properties have "cpu: 2". I assume that this should be "cpu: 1". Both the vSphere and vCloud examples have the same description and cloud properties for the example.


chipchilders:
CLA has been received. This PR is cleared to be accepted if appropriate.

PR #98, opened 2016-03-01
Update s3-release-blobstore

dpb587:
I've heard the `folder`-related question come up in Slack a couple times.


cppforlife:
`folder` option will eventually be removed because we are moving towards using s3cli everywhere.

PR #99, opened 2016-03-01
Add self signed endpoints

voelzmo:


openstack-self-signed-endpoints.html.md.erb
```diff
@@ -0,0 +1,29 @@
+---
+title: Validating self-signed OpenStack endpoints
+---
+
+<p class="note">Note: This feature is available with bosh-openstack-cpi v23+.</p>
+
+When your OpenStack is using with a self-signed certificate, you want to enable the OpenStack CPI to validate it. You can configure the OpenStack CPI with the public certificate of the RootCA that signed the OpenStack endpoint certificate.
```
dpb587-pivotal: I think "is using with a" is meant to be "is using a" or "is running with".

voelzmo: Awesome, thanks for proof-reading @dpb587-pivotal

PR #101, opened 2016-03-04
Fix bosh ssh example

wlindner:
`bosh redis ssh ls -R` -> `bosh ssh redis 0 "ls -R"`


wlindner:
Looks like #100 covers this change and more. Closing.

PR #107, opened 2016-03-16
Reword cloud-config opt-in warning

gberche-orange:
Most of the release manifest generation don't yet support cloud-config  (such as cf-release). Opt-in would prevent deploying them unless their manifest generation gets patched.

See related slack discussion at https://cloudfoundry.slack.com/archives/bosh-users/p1458080218000039


cppforlife:
Merged.

PR #109, opened 2016-03-23
Update create-release.html.md.erb

dpb587-pivotal:


dpb587-pivotal:
No write access for direct edits

PR #111, opened 2016-03-29
downcase Step

7hunderbird:
As I was working through the text learning about building a Bosh Release I found that this could be down-cased to look correct.

Thanks for all the great docs!


cppforlife:
Updated.

PR #118, opened 2016-04-20
Clarify between Availability Zone and Region with examples.

mbaltrusitis:
Give examples between Availability Zone and Region and explicitly state that REGION should be updated; following suit with the other fields in the guide.


cppforlife:
Added examples: https://github.com/cloudfoundry/docs-bosh/commit/2bae267db86e3ab67627f2a0023f538af1782878.

PR #122, opened 2016-04-21
Update Azure docs for Azure CPI v10.

AbelHu:


cppforlife:
Not sure why CLA popup is here. AbelHu is a CF contributor for Azure CPI.

PR #141, opened 2016-05-12
Add note that nova-generated keys are unusable

voelzmo:
workaround: generate your own key and import it


dpb587-pivotal:
hey cool, an official doc to refer people to :)

voelzmo:
@dpb587-pivotal :P

PR #144, opened 2016-05-18
Update problems.html.md.erb

jadekler:
The bolded text doesn't show up inside the quote (it gets `strong: 500` which is equivalent to the surrounding text), which is a little weird. I'm using Chromium `Version 49.0.2623.108 Ubuntu 14.04 (64-bit)`. Strangely, in a gist (https://gist.github.com/jadekler/53b81d623f8c0e8f3bf0a66a1269b88c) it works fine.

Simple solution, per the PR, is just to add some colons.

### DIFF
--- a/problems.html.md.erb
+++ b/problems.html.md.erb
@@ -10,19 +10,19 @@ Developer/operator communities have come far in solving similar situations with 
 
 BOSH was purposefully constructed to address the four principles of modern [Release Engineering](http://en.wikipedia.org/wiki/Release_engineering) in the following ways:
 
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
+1 - I think the colons make it more readable.

I'm not sure how the addition of colons would actually affect the CSS styling though - seems like separate parsing vs styling scopes. Formatting and styling on bosh.io looks consistent with the current Markdown.

dpb587-pivotal:
Also, looks like you're a Pivot. If you publish your Pivotal org membership, cfdreddbot won't nag you quite so much next time.

PR #148, opened 2016-05-26
Add operator

7hunderbird:
Seeing "operator" over and over again caused me to have to search for it's meaning.  And I'd like to see it on the terminology page.  Feel free to add/refactor definition which was obtained [here](https://bosh.io/docs/bosh-components.html).


cppforlife:
Added in https://github.com/cloudfoundry/docs-bosh/commit/e840a1f5d6ae7a3ea942ab1ba8f14f6a5da659f7.

Not sure why CLA thing is popping up for you...

PR #152, opened 2016-06-14
Update docs to reflect required key-value for az.cloud_properties.datacenters.clusters

DennisDenuto:
- The docs are wrong in this case, causing an error when #values gets called on a string.

Signed-off-by: Alvaro Perez-Shirley <aperezshirley@pivotal.io>


cppforlife:
Investigating if we regressed in configurations here. Thanks for noticing @DennisDenuto

PR #159, opened 2016-06-28
`gem install` docs use non-deprecated flags

wendorf:
--no-ri and --no-rdoc are deprecated in favor of --no-document

### DIFF
--- a/bosh-cli.html.md.erb
+++ b/bosh-cli.html.md.erb
@@ -5,7 +5,7 @@ title: BOSH Command Line Interface
 BOSH Command Line Interface (CLI) is used to interact with the Director. The CLI is written in Ruby and is distributed via `bosh_cli` gem.
 
 <pre class="terminal">
-$ gem install bosh_cli --no-ri --no-rdoc
+$ gem install bosh_cli --no-document
 </pre>
 
 <p class="note">Note: BOSH CLI requires Ruby 2+</p>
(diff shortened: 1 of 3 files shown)


dpb587-pivotal:
Thanks! If we're still supporting/recommending ruby 1.9.3 in the docs or elsewhere, we may not want to merge this. On our dev workstation, at least, older `gem` doesn't support `--no-document`...

    $ gem install bosh_cli --no-document
    ERROR:  While executing gem ... (OptionParser::InvalidOption)
        invalid option: --no-document
    
    $ ruby -v
    ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin14.5.0]
    
    $ gem -v
    1.8.23.2

I don't know ruby/gems all that well though.

wendorf:
Ruby 1.9.3 has been out of official support from the Ruby side since [February 2015](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/), and Ruby 2.0.0 has been out of official support since [February 2016](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/). That makes me lean toward suggesting not supporting them, but I don't know what other considerations you might have.

cppforlife:
Unfortunately we still support them for variety of users. Closing.

PR #160, opened 2016-06-29
Fixed documentation based on #112359727

joek:
The example is wrong. After long research I found the correct one base on https://www.pivotaltracker.com/n/projects/956238/stories/112359727 Please verify with the team first, but actually now it is working in our deployment.


cppforlife:
Thanks!

PR #163, opened 2016-07-01
Fix underscore formatting

voelzmo:


dpb587-pivotal:
Good change. Perhaps a more general question/discussion I've wondered about... should all properties/job/package names be backtick-quoted like this? Both for consistency and to explicitly exemplify tokens which are "technically"-sourced.

voelzmo:
@dpb587 I'm all for it, sounds good.

PR #169, opened 2016-07-13
Update description for Azure disk size and ephemeral disk size

AbelHu:


azure-cpi.html.md.erb
```diff
@@ -115,7 +117,7 @@ Example of 10GB disk:
 ```yaml
 disk_pools:
 - name: default
-  disk_size: 10_240
+  disk_size: 10_240 # On Azure the disk size must be in the range [1 * 1024, 1023 * 1024].
```
cppforlife: seems like we already specify this above. no need to repeat.

AbelHu: Above only mentions for the size of ephemeral_disk. And it does not mention that the size must be larger than 1024 MB. So we think we should add a comment here.

cppforlife: lets add that info above then so that its all in one place.

AbelHu: OK. But disk_size is not a property of cloud_properties, should we put it above "Schema for `cloud_properties` section:"
```
---
## <a id="disk-pools"></a> Disk Pools / Disk Types

* **disk_size** [Integer, required]: Size of the disk in megabytes. On Azure the disk size must be in the range [1 * 1024, 1023 * 1024].

Schema for `cloud_properties` section:

* **caching** [String, optional]: Type of the disk caching. It can be either `None`, `ReadOnly` or `ReadWrite`. Default is `None`. Only `None` and `ReadOnly` are supported for premium disks.

```

cppforlife: may be lets put it just above as in your example.

AbelHu: How about as below?

Example of 10GB disk:

* **disk_size** [Integer, required]: Size of the disk in megabytes. On Azure the disk size must be in the range [1 * 1024, 1023 * 1024].

```yaml
disk_pools:
- name: default
  disk_size: 10_240
```

YuPengZTE:
@AbelHu 请教一下，Azure这个不是PasS平台吗？看你们的讨论好像是当做IaaS与Cloud Foundry结合的？是这样的吗？谢谢！

AbelHu:
@YuPengZTE Here may not be the right place to discuss your question. You can send your question to my email chou.hu@microsoft.com. I will reply you.

PR #184, opened 2016-08-05
Change the cert generation script to support DNS name

Kaixiang:
Signed-off-by: Derek Reeve dreeve@pivotal.io

Hi:

After we enabled bosh uaa with the certs generated with the original script, we couldn't target it with DNS name.

We made some modifications to the script so that it support both DNS and ip address.

Could you help review it, with many thanks.


cppforlife:
Rewritten to use CLI v2 to generate certs.

PR #194, opened 2016-08-18
Mispelling?

skibum55:
I believe you meant to say _now_ instead of _not_ based on the remainder of the sentence


cppforlife:
Fixed. Thanks.

PR #196, opened 2016-08-24
more elaborate director_certs.sh script which generates sample template

drnic:


director-certs.html.md.erb
```diff
@@ -8,16 +8,27 @@ Depending on you configuration, there are up to three endpoints to be secured us
...
 
-certs=`dirname $0`/certs
+ip=$1
+if [[ "${ip}X" == "X" ]]; then
+  echo "USAGE: ./director_certs.sh [public IP | path to bosh-init manifest]"
+  exit 1
+fi
+if [[ -f ${ip} ]]; then
+  ip=$(cat ${ip} | yaml2json | jq -r ".jobs[] | select(.templates[].name | contains(\"director\")).networks[0].static_ips[0]")
+fi
```
voelzmo: Can we just have the IP as user input instead of doing this cat-yaml2json-jq-spruce stuff?

director-certs.html.md.erb
```diff
@@ -56,18 +67,70 @@ EOL
...
+}
+
+
+cat > certs-${ip}.yml <<EOF
+---
+properties:
+  director:
+    ssl:
+      key: |
+$(cat director.key | indent)
```
voelzmo: Instead of having a hierarchy (`director.ssl.key`), those should probably be flat values so they can be merged in by the new golang-cli?

cppforlife:
Rewrote director creds page to use CLI v2 so that it's easier to generate certs.

PR #208, opened 2016-09-08
Don't change ownership recursively in startup script

voelzmo:
pre-start might already have written some logs, we shouldn't change the ownership.

### DIFF
--- a/create-release.html.md.erb
+++ b/create-release.html.md.erb
@@ -131,7 +131,7 @@ case $1 in
 
   start)
     mkdir -p $RUN_DIR $LOG_DIR
-    chown -R vcap:vcap $RUN_DIR $LOG_DIR
+    chown vcap:vcap $RUN_DIR $LOG_DIR
 
     echo $$ > $PIDFILE
 


create-release.html.md.erb
```diff
@@ -131,7 +131,7 @@ case $1 in
 
   start)
     mkdir -p $RUN_DIR $LOG_DIR
-    chown -R vcap:vcap $RUN_DIR $LOG_DIR
+    chown vcap:vcap $RUN_DIR $LOG_DIR
```
dpb587-pivotal: Unless a release wants a more restrictive/open permissions than the default agent creates?

voelzmo: Doing -R in the startup script has some side-effects that (probably) most users don't think about when copying this from bosh.io. So while some users might actually want that (or a similar) side-effect, I'd say let's optimize for the standard user. Wdyt?

dpb587-pivotal: Makes sense

cppforlife:
I think by default it should be fine to use -R

PR #210, opened 2016-09-13
Azure CPI use env.bosh.group instead of env.bosh.group_name

gossion:
This doc is change for azure cpi https://github.com/cloudfoundry-incubator/bosh-azure-cpi-release/issues/205


dpb587-pivotal:
Thanks - changes made via #215 instead.

PR #211, opened 2016-09-14
Fix HTML typo

svrc-pivotal:
The page formatting is currently bugged due to a closing a-tag instead of a p-tag after the note, see https://bosh.io/docs/director-api-v1.html


dpb587-pivotal:
Merged - thanks!

PR #213, opened 2016-09-20
Typo/Formatting fixes

johnsonj:
comiled-releases: bad link, can't use markdown inside of html block
director-api-v1: missing end of </p> block messes up formatting page
uploading-releases: https:http should be https:


cppforlife:
Rebased and merged.

PR #214, opened 2016-09-22
Use consistent markdown for link names

dreeve:
There was a formatting issue in the [links documentation](https://bosh.io/docs/links.html). This will hopefully fix it!

(As an aside, if "web" needs to be changed I can do that as well and update this pull request)

### DIFF
--- a/links.html.md.erb
+++ b/links.html.md.erb
@@ -4,7 +4,7 @@ title: Links
 
 <p class="note">Note: This feature is available with bosh-release v255.5+.</p>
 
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Nice! Let's do the job and network names, too, like you suggest. Consistency FTW.

Also, why are you trying to confuse @cfdreddbot with your lack of public orgs-ness? I'm pretty sure I've seen you around the office and can trust the licensing of your PRs...

dreeve:
@dpb587-pivotal Yikes, good call! I've made the org public and updated the pull request to include the network names.

I'll also do as @cfdreddbot suggested and close/open the pull request after making the org public. Let me know if I should make another pull request. Thank you!

PR #216, opened 2016-09-25
fix link

pivotal-sukhil-suresh:


dpb587-pivotal:
Reminder to publish a Pivotal GitHub org so @cfdreddbot isn't so noisy.

PR #218, opened 2016-10-03
Wrong closing tag

pivotal-danjahner:


dpb587-pivotal:
Fixed via #211 - thanks though! It'll take a bit for bosh.io to get updated.

Also, a reminder to publish a Pivotal GitHub org membership so @cfdreddbot doesn't get so cranky about the CLA in the future.

PR #219, opened 2016-10-04
BOSH Documentation for vSphere NSX integration

cunnie:
- documents how to configure NSX Security Groups in Director's manifest and in resource pool
- documents how to configure NSX LBs
- fixed `default_disk_type` malformatting

[#127508577](https://www.pivotaltracker.com/story/show/127508577)

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>

### DIFF
--- a/vsphere-cpi.html.md.erb
+++ b/vsphere-cpi.html.md.erb
@@ -11,7 +11,7 @@ Schema for `cloud_properties` section:
 * **datacenters** [Array, optional]: Array of datacenters to use for VM placement. Must have only one and it must match datacenter configured in global CPI options.
     * **name** [String, required]: Datacenter name.
     * **clusters** [Array, required]: Array of clusters to use for VM placement.
-        * **\<cluster name\>** [String, required]: Cluster name.
+        * **&lt;cluster name&gt;** [String, required]: Cluster name.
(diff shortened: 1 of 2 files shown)


vsphere-cpi.html.md.erb
```diff
@@ -58,7 +58,7 @@ networks:
 vSphere CPI does not support dynamic or vip networks.
 
 ---
-## <a id='resource-pools'></a> Resource Pools / VM Types
+## <a id='resource-pools'></a>VM Types (Resource Pools)
```
dpb587-pivotal: Since this is a header used across all CPIs, it would be nice if this were updated elsewhere. Or saved for a followup commit?

cunnie: We reverted the Resource Pool to VM Type change to more consistent with the rest of the docs.

cunnie:
Force pushed to switch from Security Tags to Security Groups.

cppforlife:
Looks like vsphere-nsx.html is missing. Do we need it?

cunnie:
@cppforlife We don't  need`vsphere-nsx.html`. IIRC, we originally planned to break-out the NSX stuff in a page separate from `vsphere-cpi.html`, but on further consideration it made more sense to embed the NSX information into the vSphere CPI page.

PR #221, opened 2016-10-05
Specify host for S3 compatible blobstore

voelzmo:
From https://github.com/cloudfoundry/bosh/issues/1455 it seems this isn't easy to find out yourself.


dpb587-pivotal:
Thanks!

PR #230, opened 2016-10-24
Add instructions for recovering from a lost persistent disk

cunnie:
- Only covers deleting the disk reference, recovering data was out of scope

Signed-off-by: Brian Cunnie bcunnie@pivotal.io

[#132871917](https://www.pivotaltracker.com/story/show/132871917)


cppforlife:
@cunnie is there a bug with `bosh deploy`?

PR #231, opened 2016-10-25
Add document for assign_dynamic_public_ip in Azure CPI.

gossion:


cppforlife:
@gossion for future CPI descriptions please add release version (eg `available in v18+`) so that it's easier to know when features are available.

PR #233, opened 2016-10-26
Adding disaster recovery in index

gberche-orange:
To make this content available since 2014 more discoverable from the bosh.io/docs entry point


cppforlife:
@gberche-orange all the content in that page is already covered by other pages. also it's not really disaster recovery related, but just usual bosh health management. will remove that page from the repo.

gberche-orange:
ok thanks Dmitry for the update.

PR #244, opened 2016-11-16
Fix incorrect cloud config statement

zmb3:
Director 257+ allows for both V1 and V2 manifests.

### DIFF
--- a/manifest-v2.html.md.erb
+++ b/manifest-v2.html.md.erb
@@ -4,7 +4,7 @@ title: Manifest v2 Schema
 
 <p class="note">Note: This feature is available with bosh-release v255.4+.</p>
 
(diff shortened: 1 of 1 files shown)


zmb3:
Should be covered under the Pivotal CLA, right?

dpb587-pivotal:
Should be fine. If you publish one of your Pivotal/CF organizations to your profile, @cfdreddbot will be a little less annoyed :)

zmb3:
Is being in the **pivotalservices** org not sufficient?

image

dpb587-pivotal:
You're a member, but they're not actually public - just visible to you when logged in. Visit the individual orgs and go to the People tab to find yourself and the option to make your membership public. The public membership is what @cfdreddbot goes off of.

zmb3:
Ah, got it! Thanks for the pointer.  I made them public now.

dpb587-pivotal:
@cfdreddbot agrees - thanks!

PR #247, opened 2016-11-23
Add docs for nova-networking compatibility

voelzmo:
As of release v28, neutron is the default.

### DIFF
--- a/openstack-nova-networking.html.md.erb
+++ b/openstack-nova-networking.html.md.erb
@@ -1,0 +1,28 @@
+---
+title: Using nova-networking
+---
+
+<p class="note">Note: This feature is available with bosh-openstack-cpi v28+.</p>
+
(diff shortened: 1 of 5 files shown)


openstack-nova-networking.html.md.erb
```diff
@@ -0,0 +1,28 @@
...
+---
+
+<p class="note">Note: This feature is available with bosh-openstack-cpi v28+.</p>
+
+The OpenStack CPI v28+ uses neutron networking by default. This document describes how to enable nova-networking instead if your OpenStack installation doesn't provide neutron. **Note:** nova-networking is deprecated as of the OpenStack Newton release and will be removed in the future.
+
+1. Configure OpenStack CPI
+
+    Modify [the default template](init-openstack.html#create-manifest). In `properties.openstack`:
+    - add property `use_nova_networking: true`
```
dpb587: Soon maybe we can use inline ops syntax to recommend these sorts of changes to people. Would have the benefit of being mostly human-readable and copy-paste-able. Might be nice to default `type` to `replace` in go-patch if going that direction, which I think I'm coming around to. Or, directly embed the ops file from a repo so it has a `wget`-able link, too.

PR #257, opened 2016-12-14
BLOCKED: Document usage of `cpu`, `ram`, and `ephemeral_disk_size` as top-level `vm_type` properties

ljfranklin:
- NOTE: this should not be merged until BOSH Director and the majority of CPIs support the calculate_vm_cloud_properties call

[#126023557](https://www.pivotaltracker.com/story/show/126023557)

### DIFF
--- a/cloud-config.html.md.erb
+++ b/cloud-config.html.md.erb
@@ -55,13 +55,13 @@ azs:
 
 vm_types:
 - name: small
-  cloud_properties:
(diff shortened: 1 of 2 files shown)


cppforlife:
vm/cpu/ephemeral_disk_size wont be represented through vm types but rather go onto an instance group since this is not iaas specific.

PR #259, opened 2016-12-15
Bosh links overview

mboldt:
@cholick and I had a hard time grokking the bosh links documentation. We think this addition of a brief "Overview" section with a small, complete example, which clearly shows what goes where, what is required, what is optional, and how it's used will help others understand bosh links more quickly.

### DIFF
--- a/links.html.md.erb
+++ b/links.html.md.erb
@@ -56,7 +56,7 @@ export DATABASE_PORT="<%= link('db_conn').p('port') %>"
 # ...
 ```
 
-If the application uses two database connections, each provided by a separate database_job instance, the link becomes ambiguous and we must resolve the ambiguity in the deployment manifest. In this case, the application job's spec file could look like:
+If the application uses two database connections, each provided by a separate database instance group, the link becomes ambiguous and we must resolve the ambiguity in the deployment manifest. In this case, the application job's spec file could look like:
 
 ```yaml
 # Application job spec file.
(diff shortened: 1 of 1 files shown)


links.html.md.erb
```diff
@@ -9,6 +9,96 @@ Previously, if network communication was required between jobs, release authors
...
+  type: conn
+```
+
+Then, in the application job's templates, it can use the connection information from the link:
+
+```sh
+#!/bin/bash
+# Application's templated control script.
+# ...
+export DATABASE_HOST="<%= link('db_conn').address %>"
```
cppforlife: should be instances[0].address

mboldt: Thank you, I've fixed this.

links.html.md.erb
```diff
@@ -9,6 +9,96 @@ Previously, if network communication was required between jobs, release authors
...
+export DATABASE_PORT="<%= link('db_conn').p('port') %>"
+# ...
+```
+
+If the application uses two database connections, each provided by a separate database_job instance, the link becomes ambiguous and we must resolve the ambiguity in the deployment manifest. In this case, the application job's spec file could look like:
```
cppforlife: ah. oops misunderstood.

cppforlife:
@mboldt left a few comments. otherwise looks good.

links.html.md.erb
```diff
@@ -9,6 +9,97 @@ Previously, if network communication was required between jobs, release authors
...
+
+And then in the deployment manifest, we disambiguate the links with `provides` and `consumes` declarations on the jobs. The database jobs name their `database_conn` links. The application job uses these names to specify which database it will use for each of its links.
+
+```yaml
+# ...
+instance_groups:
+- name: database_1
+  jobs:
+  - name: database_job_1
+    templates:
```
mboldt: Sorry, I'm not following. What should be present? What is already represented by a job in an instance group? My understanding is that the `templates` sections need to be here to map each of the instance group's jobs to a job template in a release.

mboldt:
Thank you for the feedback and corrections, @cppforlife. Much appreciated! I hope this addition will make bosh links more accessible to users.

PR #261, opened 2016-12-20
No package yajl-ruby available in CentOS.

nunodio:
The package yajl-ruby isn't available in CentOS. it can be installed in the end with:
```
gem install yajl-ruby
```

### DIFF
--- a/install-bosh-init.html.md.erb
+++ b/install-bosh-init.html.md.erb
@@ -41,7 +41,8 @@ Here are the latest binaries:
 	**CentOS**
 
 	<pre class="terminal">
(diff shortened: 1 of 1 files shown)


d:
What's the problem you're trying to fix? Nothing seems to break without manually installing this gem

nunodio:
The package yajl-ruby isn't available in CentOS. 
When you do :

```
yum install yajl-ruby
```

the package is not found, so you need to install it with _gem._
I only use the official CentOS repository. Do you use additional repositories like rpmfusion etc?

d:
I don't think `yajl-ruby` needs to be installed anyway. It wasn't probably needed ...

nunodio:
If is not needed it, probably the best move is to remove it from the dependencies, in order to keep the documentation updated. Can you confirm that?

nunodio:
@d do you have any new information about this topic?

PR #268, opened 2017-01-04
Add example for director ssl configuration

Kiemes:
Director manifest cannot use file path for certificates and keys.
The documentation suggests otherwise.


Kiemes:
@cppforlife thx for merging. The [formatting](http://bosh.io/docs/director-certs.html) somehow is messed up. Do you need an empty line before ```yaml?

PR #273, opened 2017-01-16
Add info on AWS lb_target_groups cloud property

ljfranklin:
* AWS CPI now supports attaching instances to Load Balancer Target
Groups. This is necessary for VMs to be attached to Application Load
Balancers.

[#135963933](https://www.pivotaltracker.com/story/show/135963933)

Signed-off-by: Brian Cunnie <bcunnie@pivotal.io>


ljfranklin:
Gated on AWS CPI v63 release

PR #275, opened 2017-01-23
Add cross-links for Softlayer CPI

petergtz:
/cc @suhlig


suhlig:
Hi @cppforlife, is there an outlook when the docs that contain this commit will be published? We'd like to deliver the [associated story](https://www.pivotaltracker.com/story/show/137784451) ;-)

PR #283, opened 2017-02-27
Correctly identify context ID as always being a string, never null

winnab:
Documentation added in https://github.com/cloudfoundry/docs-bosh/pull/280 incorrectly indicated that context_id could be returned as a null value. The actual implementation will always return a string, an empty one if context ID was not set. This pull request corrects that error.

[#140016781]


winnab:
Hi team--I've publicised my Pivotal Github org, so hopefully this should now be good to go. Please let me know if there is anything else I need to do. Thanks! --Winna

PR #293, opened 2017-04-05
Update build-stemcell.html.md.erb

pnikonowicz:
Fix broken hyperlink


dpb587-pivotal:
Thanks for making a PR and mentioning this! Since there are some more specific links which can be used instead, I went ahead and updated this link and a couple others in that file in [b0b2cdac5](https://github.com/cloudfoundry/docs-bosh/commit/b0b2cdac50b5d45fa1be0934fe9af5896a1fb8ac).

PR #304, opened 2017-05-19
URL fixes for bad links

keitwb:


cppforlife:
Fixed on master. Thanks.

PR #307, opened 2017-05-29
Include sha2 flag in create-release README

henderjm:


henderjm:
Hey, I just publicised my membership on the pivotal-cf org. I wasn't aware it was set to private at the time of opening this PR.

PR #308, opened 2017-05-30
Update vsphere-esxi-host-failure.html.md.erb

eamonryan:
Corrected language pertaining to what vSphere HA actually does in the event of a host failure.

vSphere HA does not recreate VMs because it does not actually destroy them.

It unregisters the VMs from the failed host and then registers VMs on good hosts - no VM files are deleted. The process is just started on another host using the same exact config files and virtual disk files which reside on shared storage.


eamonryan:
I have fixed my membership visibility (pivotal-cf) - can someone recheck?

PR #321, opened 2017-07-02
Update cli-envs.html.md.erb

stevewallone:
Added the option of using BOSH_ENVIRONMENT to specify the bosh environment.


cppforlife:
Fixed on master.

PR #322, opened 2017-07-06
Document BOSH_LOG_LEVEL environment variable

gberche-orange:
So that users can discover how to turn on verbose logging. 

See https://github.com/cloudfoundry/bosh-cli/issues/113 and https://github.com/cloudfoundry/bosh-utils/blob/f8bb60f0bb8b25aeeafe837e5e851ec3d7d850e0/logger/logger.go#L24-L30

### DIFF
--- a/cli-v2.html.md.erb
+++ b/cli-v2.html.md.erb
@@ -922,6 +922,10 @@ See [CPI config](cpi-config.html).
 ---
 ### <a id="misc"></a> Misc
 
+- <a id="debug"></a> `BOSH_LOG_LEVEL=debug bosh ...`
+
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Thanks! I think this is already documented in [docs/cli-global-flags](http://bosh.io/docs/cli-global-flags.html#logging), but that page is not very discoverable (others have also overlooked this page).

Perhaps we can keep this PR open until @cppforlife highlights that globals page a bit more so everyone can find it. At the very least, linked from [cli-v2](http://bosh.io/docs/cli-v2).

cppforlife:
Fixed by d90c975955a2815863d2dd2bdabe15e82de6e055.

gberche-orange:
thanks @dpb587-pivotal and @cppforlife

PR #326, opened 2017-07-10
Command `bosh status` doesn't work in V2, change to `bosh env`

kdvolder:


dpb587-pivotal:
For what it's worth, the new CLI does not require the `director_uuid` property to be present in the manifest at all :)

kdvolder:
Hmmm... well, that's not what the doc says. It says its required in the V2 schema. It would be nice if someone could clarify if / when it is actually required.

I was guessing... it *is* required if you use V2 manifest schema, but not required if your manifest looks like its still using V1 schema. But really a guess is all it is. Not finding the docs very clear on this.

I'd fix em in a PR, but I think I'm too confused right now to do anything more but suggest some basic edits like `cf status` -> `cf env`.

dpb587-pivotal:
Yeah, docs should be updated one way or another. You definitely can leave out `director_uuid` if you're using the CLI v2.

Comment was more an FYI for you+others, but also a reminder to the PM to think about what the docs should say to be most correct.

Appreciative for the PRs, regardless - thanks!

dpb587-pivotal:
Closing - the text was updated to indicate UUID is not required by CLI v2, so going to leave the old CLI v1 command there.

PR #333, opened 2017-07-19
Removing the reference to bookbinder

seviet:


seviet:
whoops! my CF org membership was not public, but now it is!

dpb587-pivotal:
Confirmed - thanks!

Note to self - this as a "book" was removed in cloudfoundry/docs-book-cloudfoundry@0a78cd4297d9d70360c61de441e18c7991fe6800.

@cppforlife, you should add https://bosh.io/docs as the website for this repo.

PR #334, opened 2017-07-20
Fix command line continuation in docs

dgkimura:


dpb587-pivotal:
Thanks! Also, reminder to make your Pivotal or CF org membership public so @cfdreddbot doesn't fret next time.

PR #339, opened 2017-07-28
Fix blobstore private.yml example

henryaj:
Seems like the format for `private.yml` changed at some point.


cppforlife:
@henryaj as noted on the top the doc included instructions apply to cli v2 which expects blobstore.options instead of blobstore.s3.

PR #346, opened 2017-08-06
Enhance openstack init documentation

hoegaarden:
- Link bosh variables with
  - terraform outputs
  - Openstack RC files / variables
- Add other minor hints and tips

### DIFF
--- a/init-openstack.html.md.erb
+++ b/init-openstack.html.md.erb
@@ -151,33 +151,38 @@ You must create and configure two Security Groups to restrict incoming network t
     $ git clone https://github.com/cloudfoundry/bosh-deployment
 
     # Fill below variables (replace example values) and deploy the Director
-    $ DIRECTOR_NAME='bosh-1'     # arbitrary name for the director
-    $ NET_INT_CIDR='10.0.0.0/24' # terraform output 'network CIDR'
-    $ NET_INT_GW='10.0.0.1'      # terraform output 'network gateway IP'
-    $ NET_INT_IP='10.0.0.6'      # terraform output 'director private ip'
-    $ NET_INT_ID='c88c7cdc-4...' # terraform output 'network id'
-    $ DEFAULT_KEY_NAME='bosh'    # terraform output 'keypair name'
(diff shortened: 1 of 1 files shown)


hoegaarden:
CLA signed and sent via email.

hoegaarden:
CLA has been accepted and filed, everything should now be set up to allow my contributions.
Let me know if there is anything else I need to do.

init-openstack.html.md.erb
```diff
@@ -151,28 +151,36 @@ You must create and configure two Security Groups to restrict incoming network t
     $ git clone https://github.com/cloudfoundry/bosh-deployment
 
     # Fill below variables (replace example values) and deploy the Director
+    $ DIRECTOR_NAME='bosh-1'     # arbitrary name for the director
```
cppforlife: how about putting comments inline? i think extra indirection thru vars is not really helpful.

hoegaarden: I wanted to do that initially. However, when breaking long lines with `\` one cannot add comments afterwards. Also, I wanted to keep the code sample c&p-able.

One way I see is e.g.:
```
command \
    # a comment on arg \
    -some arg \
    # a comment on other_arg \
    -some other_arg
```

Just let me know what you'd prefer and I can adopt my change.

hoegaarden: One thing I could do (which I personally do not like that much):
```
command \
    `# --first is very important` \
    --first arg \
    `# --second is super important too` \
    --second arg
```
Any better way to do inline comments properly?

cppforlife: @hoegaarden how about this format: https://github.com/cloudfoundry/docs-bosh/blob/master/init-vsphere.html.md.erb#L43 ?

hoegaarden: @cppforlife Sure, I'll rework the PR as soon as possible. Thanks for your input.

hoegaarden:
@cppforlife does that look ok to you?

cppforlife:
@hoegaarden just built https://github.com/cppforlife/js-annotate to make explanations a bit more easier to read. will try to use it for this page.

PR #348, opened 2017-08-08
update v2 status command

7hunderbird:


dpb587-pivotal:
Hi - I think these docs are assuming the user has set either `BOSH_ENVIRONMENT` or implicitly including `-e my-env` (given the note after the table of "Most commands require (`--environment`) `-e` and `--deployment` (`-d`) flags").

To keep this a more concise and readable table, I'd propose not merging this.

Is there something particular that led to confusion on this line? Perhaps there's a different change we should make?

7hunderbird:
Yea sorry, I didn't provide any context in my PR, apologies.

I agree it's more readable as is.  It just didn't work for me and I was hoping it would because of how straight forward `bosh status` was to use in the bosh-cli v1.

Going forward though I think it's alright to keep as is.

PR #350, opened 2017-08-09
Updated blobstore credentials format

cagiti:


cppforlife:
@cagiti i just removed reference/ since it was out of date. same info is available in http://bosh.io/docs/release-blobstore#s3-config

PR #351, opened 2017-08-10
Add `set_disk_metadata` to CPI API methods

voelzmo:
[#141967195](https://www.pivotaltracker.com/story/show/141967195)

Signed-off-by: Marco Voelz <marco.voelz@sap.com>

### DIFF
--- a/cpi-api-v1.html.md.erb
+++ b/cpi-api-v1.html.md.erb
@@ -502,6 +502,8 @@ No return value
 ---
 ### <a id="set-disk-metadata"></a> `set_disk_metadata`
 
+<p class="note">Note: This method is called by BOSH v262+.</p>
+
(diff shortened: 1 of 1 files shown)


cpi-api-v1.html.md.erb
```diff
@@ -500,6 +500,44 @@ No return value
...
+  }
+]
+
+```
+
+#### Returned
+
+No return value
+
+[Example #set\_disk\_metadata](https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release/blob/0c8ee8951cab41d0ddc86591719d55d8a783ac98/src/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb#L629)
```
dpb587-pivotal: :+1: canonical link

PR #358, opened 2017-08-15
Add docs on service_account property

akshaymankar:


akshaymankar:
There seem to be a few more properties which are not documented. 
Reference: https://github.com/cloudfoundry-incubator/bosh-google-cpi-release/blob/b32fcf7e94d390bd6e340b1e07fd9f4899370b51/src/bosh-google-cpi/action/cloud_properties.go#L42

dpb587-pivotal:
Thanks - going to close this as it's superseded by #401 from the Google CPI team which documents the other properties you mentioned as well.

Also, reminder to publish one of your Pivotal/CF GitHub org memberships so @cfdreddbot is a bit more friendly next time.

akshaymankar:
Thanks @dpb587 :)

PR #362, opened 2017-08-22
Update init-vsphere.html.md.erb

stevewallcgi:
Changed to use the BOSH_ENVIRONMENT variable to alleviate the need for the '-e bosh-1' option on the command line.


cppforlife:
canonical form is to specify `-e` to avoid any confusion. may be in practice once user is familiar they do end up using env var but i think it makes sense to keep it more explicit in the first steps.

PR #365, opened 2017-08-25
Updated backend-service cloud properties usage

cagiti:


cppforlife:
@cagiti care to add an example?

cagiti:
@cppforlife I've added configuration examples, let me know if they need further work. Cheers 👍

PR #382, opened 2017-09-28
Update GCS blobstore documentation.

bengtrj:


release-blobstore.html.md.erb
```diff
@@ -39,6 +39,22 @@ blobstore:
...
+
+```yaml
+---
+blobstore:
+  provider: gcs
+  options:
+    bucket_name: <bucket_name>
+```
+
+You have to be logged in using gcloud to your Google account. There is no need for a `config/private.yml`.
```
dpb587-pivotal: Hi. Thanks for proposing this.

I don't think this is accurate. If users want to specify a `json_key` option, the `gcloud` commands don't need to be installed.

PR #388, opened 2017-10-13
Update deployment-manifest.html.md.erb

b10s:


cppforlife:
@b10s setting vcap password (vcap doesnt have a password by default -- cannot login) is not at this point recommended. what are you trying to use vcap user?

PR #391, opened 2017-10-15
Correctly marked jobs properties has as optional resolves #329

### LINKED ISSUE #329: `properties` in `job` in V2 manifest is not required?
The doc says that it is required here:

https://github.com/cloudfoundry/docs-bosh/blob/4ba9ee2e4129f476cafb5fbf0b52534001519e7e/manifest-v2.html.md.erb#L117

But this [sample file](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml), which I assume is a correct manifest has an example where 'properties' is not specified. So I gather that 'properties' isn't really required:

```
...
  jobs:
  - name: consul_agent
    release: consul
    consumes:
      consul_common: {from: consul_common_link}
      consul_server: nil
      consul_client: {from: consul_client_link}
  - name: nats
...
```

tracyde:


dpb587-pivotal:
Thanks!

PR #392, opened 2017-10-16
Change director_name to Bosh-Lite-Director

pivotal-sukhil-suresh:
Update command to create BOSH lite environment to use a director name without spaces; `Bosh-Lite-Director` instead of `Bosh Lite Director`. Director name with space results in cryptic error messages when using the config server. The director generated fully-qualified-variable-name will contain spaces and as a result, will fail director validation of variable names.

Ref: https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/config_server/config_server_helper.rb#L42

### DIFF
--- a/bosh-lite.html.md.erb
+++ b/bosh-lite.html.md.erb
@@ -49,7 +49,7 @@ Follow below steps to get it running on locally on VirtualBox:
       -o ~/workspace/bosh-deployment/bosh-lite-runc.yml \
       -o ~/workspace/bosh-deployment/jumpbox-user.yml \
       --vars-store ./creds.yml \
-      -v director_name="bosh-1" \
+      -v director_name="bosh-lite" \
       -v internal_ip=192.168.50.6 \
       -v internal_gw=192.168.50.1 \
       -v internal_cidr=192.168.50.0/24 \
--- a/bosh-lite.html.md.erb
+++ b/bosh-lite.html.md.erb
@@ -49,7 +49,7 @@ Follow below steps to get it running on locally on VirtualBox:
       -o ~/workspace/bosh-deployment/bosh-lite-runc.yml \
       -o ~/workspace/bosh-deployment/jumpbox-user.yml \
       --vars-store ./creds.yml \
-      -v director_name="Bosh-Lite-Director" \
+      -v director_name="bosh-1" \
       -v internal_ip=192.168.50.6 \
       -v internal_gw=192.168.50.1 \
       -v internal_cidr=192.168.50.0/24 \
--- a/bosh-lite.html.md.erb
+++ b/bosh-lite.html.md.erb
@@ -49,7 +49,7 @@ Follow below steps to get it running on locally on VirtualBox:
       -o ~/workspace/bosh-deployment/bosh-lite-runc.yml \
       -o ~/workspace/bosh-deployment/jumpbox-user.yml \
       --vars-store ./creds.yml \
-      -v director_name="Bosh Lite Director" \
+      -v director_name="Bosh-Lite-Director" \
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
This person sounds like they know a thing or two about BOSH... but @cfdreddbot doesn't quite trust them yet...

bosh-lite.html.md.erb
```diff
@@ -49,7 +49,7 @@ Follow below steps to get it running on locally on VirtualBox:
       -o ~/workspace/bosh-deployment/bosh-lite-runc.yml \
       -o ~/workspace/bosh-deployment/jumpbox-user.yml \
       --vars-store ./creds.yml \
-      -v director_name="Bosh Lite Director" \
```
dpb587-pivotal: I suppose this means secrets will be regenerated when the user next deploys.

If we're doing dashes, should we do lower case, too. And if we're changing the name... I wonder if we should remove the "director" suffix since that's kind of implied?

pivotal-sukhil-suresh: >I suppose this means secrets will be regenerated when the user next deploys.
I guess you are referring to an upgrade/update of the bosh-lite environment and it's repercussions.

> if we're doing dashes, should we do lower case, too. And if we're changing the name... I wonder if we should remove the "director" suffix since that's kind of implied?
I am happy with any name compatible with director variable naming rules :)
BTW, all the other IAAS bootstrap guidelines use the name bosh-1. Eg. http://bosh.io/docs/init-aws.html#deploy

cppforlife: this was kept this way for cf-release compatibility.

dpb587-pivotal: What cf-release compatibility? Is it okay to change it?

emalm: https://github.com/cloudfoundry/cf-release/blob/v277/scripts/generate-bosh-lite-dev-manifest#L27 comes to mind.

dpb587-pivotal: @cppforlife, okay to merge and break compatibility? I see `EXPECTED_DIRECTOR_NAME` is a thing, so there's at least a workaround.

emalm: I don't think there's any place to override that `EXPECTED_DIRECTOR_NAME` variable, though: it's not an expression like `EXPECTED_DIRECTOR_NAME=${EXPECTED_DIRECTOR_NAME:-Bosh Lite Director}`. /cc @dsabeti

dpb587-pivotal: Oh, you're right; I read that too quickly. Thanks

dsabeti: If this important to do, we could probably make that script smart enough to handle both names. It wouldn't have to be any smarter than something this (sorry for lack of bash fu):
```bash
EXPECTED_DIRECTOR_NAME_OLD="Bosh Lite Director"
EXPECTED_DIRECTOR_NAME_NEW="bosh-1"

ACTUAL_DIRECTOR_NAME="$(echo "$BOSH_STATUS" | grep Name)"

if [[ "$ACTUAL_DIRECTOR_NAME" != *"$EXPECTED_DIRECTOR_NAME_OLD"* ]] || [["$ACTUAL_DIRECTOR_NAME" != *"$EXPECTED_DIRECTOR_NAME_NEW"* ]]; then
  echo "Can only target $EXPECTED_DIRECTOR_NAME_OLD or $EXPECTED_DIRECTOR_NAME_NEW. Please use 'bosh target' before running this script."
  exit 1
fi
```

cppforlife:
fixed by ecde5dc8dc48ec0ea8725f64c17f81b66d733270.

PR #394, opened 2017-10-18
Add dashes to bosh-lite example director name

dreeve:
A director name with spaces will break the config server, if you happen
to want to deploy that. It doesn't seem like the name of this director
is used elsewhere, so we can just add dashes.

Signed-off-by: Jesse Malone <jmalone@pivotal.io>


dpb587-pivotal:
Agreed and thanks! But duplicate of the-very-recent-#392, so we can continue discussion on there :)

PR #399, opened 2017-10-24
Small corrections to cli-ops-files.html.md.erb

ansd:
```
- type: replace
  path: /items/name=item7/count
  value: 10
```

results in error (using CLI v. 2.0.44): 

> Expected to find a map key 'count' for path '/items/name=item7/count' (found map keys: 'name')

Alternatively, we could add a '?' to the end of 'count'.


tjvman:
LGTM. Merging.

Issue #398, opened 2017-10-20
Google CPI documentation does not mention `service_account` parameter


smgoller:
In the [Google CPI documentation relating to vm-types](http://bosh.io/docs/google-cpi.html#vm-types) the `service_account` parameter is not mentioned at all. This option is necessary for bosh to be able to launch an instance with the permissions of a service account. This option is described in the [bosh resource pool options section](https://github.com/cloudfoundry-incubator/bosh-google-cpi-release/blob/master/src/bosh-google-cpi/README.md#bosh-resource-pool-options) of the Google CPI README.

dpb587-pivotal:
Closing - thanks for reporting! PR #401 fixes this and the bosh.io/docs site will be updated shortly.

PR #401, opened 2017-10-26
google-cpi: properties as of v25.10.0

johnsonj:
Changed the format for this to sync with the CPI documentation. I think it's easier for the viewers and easier for us to keep the two in sync. 

fixes cloudfoundry-incubator/bosh-google-cpi-release/issues/233


google-cpi.html.md.erb
```diff
@@ -75,17 +78,25 @@ networks:
...
+| `ram`                   | Y        | Integer                                  | `2048`                                                                         | Amount of memory ([Google Compute Engine Custom Machine Types](https://cloud.google.com/custom-machine-types/)) the CPI will use when creating the instance (required if not using `machine_type`)
+| `zone`                  | N        | String                                   | `us-west1-a`                                                                   | The name of the [Google Compute Engine Zone](https://cloud.google.com/compute/docs/zones) where the instance must be created
+| `root_disk_size_gb`     | N        | Integer                                  | `10`                                                                           | The size (in Gb) of the instance root disk (default is `10Gb`)
```
dpb587-pivotal: A default of `10Gb` would suggest to me that I need to include `Gb` suffix in the value, although I think it's supposed to be just the number.

dpb587-pivotal:
Fixes #398

johnsonj:
To be transparent, I went with this format to match our CPI docs. I did merge the information on bosh.io (added the defaults column). 

IMO it's easier to use this information as a table now that the lists are quite long. I'd vote for all the CPIs moving to this format or coming up with something better. 

If that's a non starter I can roll this back and just add the few fields, but I am concerned about keeping the two lists in sync and don't want to adopt the list format in our docs.

cppforlife:
im not sure tables deal very well with nested properties (dont think google cpi has them though).

why do we maintain two set of docs? can we kill github ones?

johnsonj:
Is there a way we can embed docs from the cpi to bosh.io? Link would be so-so.

What I like about having the docs in the same project as the CPI is we can ensure as we add new features they are merged with docs. It also means the docs in the repo at a given tag represent the feature set. Separate repos complicate this.

dpb587-pivotal:
I strongly vote for the CPI repository to be the authoritative docs, and bosh.io should be able to embed those. Not the other way around. I agree with @johnsonj.

If consistent docs across CPIs would be an issue, there's always the schema option for a more structured form of documenting supported properties.

cppforlife:
hmm i could get on board with that. need to figure out though how to keep
it all consistent across different cpi repos.

dpb587-pivotal:
For the sake of getting docs updated to list the latest properties, I converted the table to the currently-typical list format. I autogenerated it from the bosh-google-cpi-release docs to retain consistency and make it predictable in the future. Eventually, I hope we can embed docs from the CPI release itself and avoid this sort of translation issue.

johnsonj:
Nice extraction code! thanks

PR #404, opened 2017-11-02
Remove erb file extension

dpb587:
* use `.md` files to support reasonable, natural rendering in editors/UIs
 * agent-cpi-interactions.html.md.erb still has some dynamic content and is trying to use `<div>` blocks; can clean up in a followup
 * `image_tag` wasn't adding anything special; should update alt tags instead of "image" in a followup
 * this may cause some conflicts in open prs
 * ran bosh-io/web locally to manually verify diff; may want to do a byte diff locally before merging if concerned (only difference should be ` alt="image"`)


cppforlife:
gonna close this one, but will drop extensions next week.

PR #417, opened 2017-11-16
Escape key for property persistent_disk_type

BenChapman:
persistent_disk_type was not escaped in the manifest-v2 schema. This change means the underscores will appear correctly.

Signed-off-by: Iain Sproat <isproat@pivotal.io>


dpb587-pivotal:
Thanks!

PR #426, opened 2017-11-28
Small fix to the --var-file example

jtuchscherer:
The file names were not matching


dpb587-pivotal:
Hi - in this particular case it's intentional. There's a difference between the `--var-file` and `--vars-file`. This example is using `--var-file` which is saying "get the entire contents of `1.txt` and set the `access_key_id` variable with it". As opposed to `--vars-file`, which is parsing a YAML file for a hash and creating variables from those keys/values.

jtuchscherer:
Makes sense. I changed it, so that the files are now consistently have the txt extension in the example.

PR #428, opened 2017-11-30
Added instructions for bosh lite on AWS and keys to ssh on to the VM.

svennela:
I have added instructions to create bosh lite on AWS.


cppforlife:
i think we need to find a better place for this since it's confusing when spinning up first director why you would need instructions for bosh-lite.

svennela:
I can clone this page, make it bosh-lite or update on below page?
http://bosh.io/docs/bosh-lite.html

dpb587-pivotal:
Closing - it's a bit confusing to introduce bosh-lite-specific details in AWS. Maybe we can make a IaaS-general section about bosh-lite at some point, but it would probably be better if it were its own section somewhere.

PR #435, opened 2017-12-10
Correct typo: flag --n should be -n

olle-brkt:
The flag for affirming any confirmation that typically requires use input, is listed as --n, when it should be -n


olle-brkt:
Signed and sent the CLA!

dpb587-pivotal:
Thanks!

PR #438, opened 2017-12-13
Add `--include-outdated` flag to configs cmd

friegger:
[#153517739](https://www.pivotaltracker.com/story/show/153517739)


cppforlife:
@friegger @voelzmo since we are going alt1 proposal for naming, sounds like we should close this?

PR #444, opened 2017-12-22
Fix documentation for SSH variable type

bandesz:
I wasn't able to validate this as I'm still using BOSH v1 at the moment but from the Concourse documentation it seems an SSH variable's public key will be generated in OpenSSH format and not PEM encoded.

If my assumption is right then please review and merge this PR.


bandesz:
I publicised my membership in Government Digital Service which should be covered under Corporate CLA.

dpb587-pivotal:
Good catch - thanks!

PR #455, opened 2018-01-29
Update bosh config documentation for new CLI args

manno:
Show config either by ID or by type and name.

### DIFF
--- a/cli-v2.html.md.erb
+++ b/cli-v2.html.md.erb
@@ -468,9 +468,9 @@ See [Configs](configs.html).
     Succeeded
     </pre>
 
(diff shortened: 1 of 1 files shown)


cli-v2.html.md.erb
```diff
@@ -481,7 +481,9 @@ See [Configs](configs.html).
     Update default `my-type` config on the Director.
 
     <pre class="terminal">
-    $ bosh -e my-env update-config my-type config.yml
+    $ bosh -e my-env config --type=my-type --name=my-name
+    # or
+    $ bosh -e my-env config 5
```
dpb587-pivotal: These look like different commands from `update-config`. Is that intentional?

PR #471, opened 2018-02-27
Escape underscores for aws_key_arn

ashwin-venkatesh:
We noticed the keys looked a little off in the docs and FTFY


dpb587-pivotal:
Thanks!

Reminder to publish your GitHub org so @cfdreddbot isn't so concerned about you next time.

PR #476, opened 2018-03-03
fixed vSphere

ktpv:


dpb587-pivotal:
Thanks! And a reminder to publish your @pivotal-cf membership so @cfdreddbot isn't quite so bothered next time.

PR #477, opened 2018-03-05
Use instance_groups instead jobs

b10s:


dpb587-pivotal:
Thanks, @b10s - I know it's a straightforward change, but would you mind signing the CLA so @cfdreddbot becomes happy about this? A signed CLA will apply to future contributions in other Cloud Foundry projects as well.

CAFxX:
@dpb587-pivotal: @b10s should be covered in our corporate CLA...

b10s:
@dpb587-pivotal sorry for late reply, I've just published Org membership. Our Org signed CLA.

PR #496, opened 2018-04-18
Consistently point create release to manifest v2

h4xnoodle:


dpb587-pivotal:
@h4xnoodle I guess you should publish a non-`pivotal-cf-experimental` membership to keep @cfdreddbot happy in the future.

And thanks!

PR #497, opened 2018-04-18
Document how the 'drain' script can catch a 'scale-in' condition

bgandon:
Hi there,

Live from User Day in Boston, here is an improvement on the `drain` script documentation.

When a node is being permanently deleted, the `drain` script sometimes has to take action so that the current node leaves its cluster gracefully.

This PR documents how to do that.

Cheers


voelzmo:
@cppforlife @dpb587-pivotal sounds like we should prioritize https://github.com/cloudfoundry/bosh-notes/blob/master/proposals/drain-actions.md when we're now actively advising people to use that workaround with `size: 0`

PR #507, opened 2018-04-26
Correct bosh generate package syntax

MirahImage:
Update `bosh generate package` to `bosh generate-package`

Consistent with the current version of the `bosh-cli` and [bosh-cli v2 docs](https://bosh.io/docs/cli-v2/#generate-package)

Paired with @blgm


dpb587-pivotal:
Thanks!

PR #516, opened 2018-05-15
Update documentation for 'local_dns.enabled'

bgandon:
Hi there,
Now that `local_dns.enabled` is `true` by default, the BOSH documentation should reflect this.
Best

### DIFF
--- a/content/dns.md
+++ b/content/dns.md
@@ -104,9 +104,14 @@ DNS release provides a way to delegate certain domains via [`handlers` property]
 ---
 ## Enabling DNS {: #enable }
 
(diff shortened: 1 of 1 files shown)


cppforlife:
default is still false in the release. it's only enabled in bosh-deployment. closing for now.

bgandon:
Sorry Dmitry, I'm bewildered here with this answer.

I personally consider that bosh-deployment is a much more important reference in the community than a default value of a job property in the BOSH Release for BOSH. I've even never had a look to job specs of this release. I'm used to trusting bosh-deployment defaults blindly... Are you saying that I should not?

The other question is: Why is Bosh DNS enabled by default in bosh-deployment and not the default in the release then?
Subsidiary question is: What's the meaning of each of these defaults?

PR #521, opened 2018-05-16
Update cck.md

pivotal-todd-robbins:
The options to enter for --recover was not documented anywhere so I am adding this here.


dpb587-pivotal:
Thanks! I reworded some things for clarity and added a warning on automating its usage. Feel free to comment if I lost some of your original intention.

PR #531, opened 2018-06-27
Update manifest to v2

ryanmattcollins:
We're just following some onboarding docs and there was a discrepancy between the v2 example and these docs which seem to be out of date. I'm not sure if this change should be made or if we should be linking to a separate v2 docs page in our onboarding.

Also unsure if all references to "job(s)" should now be references to "instance group(s)" or not.


dpb587-pivotal:
Thanks! Feel free to publish one of your GitHub org memberships to avoid @cfdreddbot complaining next time.

PR #536, opened 2018-07-17
Update the 'bosh recreate --fix' documentation

KaiHofstetter:
- Clarify the behavior of the '--fix' flag for 'bosh recreate'

[#156648106](https://www.pivotaltracker.com/story/show/156648106)

### DIFF
--- a/content/cli-v2.md
+++ b/content/cli-v2.md
@@ -662,7 +662,7 @@ See [CPI config](cpi-config.md).
     Recreates VMs for specified instances. Follows typical instance lifecycle.
 
     - `--skip-drain` flag skips running drain scripts
-    - `--fix` flag replaces unresponsive VMs
+    - `--fix` flag specifies to recover an instance with an unresponsive agent instead of erroring
     - `--canaries=` flag overrides manifest values for `canaries`
     - `--max-in-flight=` flag overrides manifest values for `max_in_flight`
     - `--dry-run` flag runs through as many operations without altering deployment


KaiHofstetter:
@dpb587-pivotal I also changed the CLI help in the related [PR](https://github.com/cloudfoundry/bosh-cli/pull/453)

voelzmo:
@dpb587-pivotal I'm not sure if this should be "instances" or "VMs". Given the output of the CLI command
```
$ bosh recreate --help
Usage:
  bosh [OPTIONS] recreate [recreate-OPTIONS] [INSTANCE-GROUP[/INSTANCE-ID]]

Recreate instance(s)
```

The entire command deals with instances rather than VMs. In the end, of course VMs are being created, but this seems to be a detail hidden by the command. Wdyt?

dpb587-pivotal:
@voelzmo yeah... I started having mixed feelings and meant to revert my change while I pondered more. I should have put my thoughts in a comment...

Considering instance is the conceptual representation and VM being the manifestation, I found it hard to imagine a concept becoming unresponsive. I was also debating because I think we currently only have messages referring to `unresponsive agent` and `unresponsive vm`, but we never show `unresponsive instance` anywhere. Thoughts?

voelzmo:
I understand your thought process and can see both sides: not having to introduce "unresponsive isntances" and sticking to "instances" in the messaging when your command claims to deal with "instances".

How about "an instance with an unresponsive agent"? Or is this already too much wording?

dpb587-pivotal:
> an instance with an unresponsive agent

Sounds good to me; I favor the verboseness for consistency in this case.

Feel free to squash my change out of history if you're good with the latest phrasing and want to commit; then we can merge it.

PR #538, opened 2018-07-20
Changed path in 'Update cloud config'

yogendra:
Path pointed to `~/workspace/bosh-env/...` instead of `~/bosh-env/...`


dpb587-pivotal:
Thanks! I shortened your corrected path a bit further.

PR #540, opened 2018-07-20
Add guide for NATS CA rotation

h4xnoodle:
Workflow for rotating NATS certificates. 

https://www.pivotaltracker.com/story/show/158608190


content/nats-certificate-rotation.md
```diff
@@ -0,0 +1,216 @@
+# Rotating NATS Certificates
+
+The following strategy rotates the certificates and CA across the director, health monitor, NATS, and all VMs. See [Components of Bosh](bosh-components.md) for more information on core components.
+
+
+![image](images/nats_rotation.png)
```
dpb587-pivotal: A first look, this image seems a little intimidating and a user probably wouldn't understand it. I do think it's helpful once they have a general understanding of the process and components, so maybe it's better at the end? Perhaps, to ensure they have a big picture summary at the start, we could instead have a bullet list with a super simple summary of steps? Something like...

1. Reconfigure VMs to trust old + new CA
1. Reconfigure director to use new CA
1. Reconfigure VMs to trust only new CA

Then the rest of the doc gets into the fun stuff which expands those conceptual steps into the implementation details.

Maybe?

content/nats-certificate-rotation.md
```diff
@@ -0,0 +1,216 @@
...
+* Each VM/agent continues to use the old certificates to communicate with NATS. 
+* 
+
+`add-new-ca.yml` 
+
+```yaml
+---
+- type: replace
+  path: /instance_groups/name=bosh/properties/nats/tls/ca?
+  value: ((nats_server_tls.ca))((nats_server_tls_2.ca))
```
dpb587-pivotal: This probably, technically, works as is. But we should probably explicitly include a newline between the values since a newline is not required at the end of a certificate (and external tooling or credhub may be doing it differently). I think?

content/nats-certificate-rotation.md
```diff
@@ -0,0 +1,216 @@
...
+      extended_key_usage:
+      - client_auth
+
+```
+
+
+### Recreate all VMs, for each deployment.
+```shell
+$ bosh -d deployment-name recreate
+```
```
dpb587-pivotal: Can we add some inline text/note/reminder about why they need to be rotated?

mkdocs.yml
```diff
@@ -134,6 +134,7 @@ pages:
       - Configuration Locations: vm-config.md
       - Using Logs: job-logs.md
       - Instance Metadata: instance-metadata.md
+    - Rotating NATS Certificates: nats-certificate-rotation.md
```
dpb587-pivotal: Note to selves; when we have more rotation docs, we may want a "Rotating Credentials" or general "Credentials" section.

content/nats-certificate-rotation.md
```diff
@@ -0,0 +1,216 @@
...
+
+* Adds new variables for the new CA/certificates/private_key. 
+* The director and health monitor are given a modified CA with the original CA and the new CA concatenated as `((nats_server_tls.ca))((nats_server_tls_2.ca))`. This allows the director to communicate with NATS using new certificates, while NATS can still communicate with the director.[1](#limitations).
+* The director and health monitor are updated to use certificates and keys generated by the new CA.
+* NATS continues to use the old certificates and private key. NATS is given the concatenated CA from above to verify certificate generated using new CA and old CA. 
+* Each VM/agent continues to use the old certificates to communicate with NATS. 
+* 
+
+`add-new-ca.yml`
```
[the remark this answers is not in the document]
h4xnoodle: I like the idea of the extraction once the docs are looking good.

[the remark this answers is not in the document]
pivotal-jamil-shamy: @dpb587-pivotal I like the idea. This brings the question of the future plan for bosh-deployment; is it going to be the official goto place for installing BOSH and other logistics ?

content/nats-ca-rotation.md
```diff
@@ -0,0 +1,225 @@
...
+ --vars-store ./creds.yml \
+ -v ... additional vars
+```
+
+`remove-old-ca.yml`
+
+* `nats.tls.ca` is updated to remove the old CA from the concatenated CAs. 
+* The director and health monitor continue to use only new certificates. The CA is updated to be just the new CA.
+* The NATS server is updated to use a new certificate and private key, generated by the new CA.
+* All components can communicate using the new CA.
```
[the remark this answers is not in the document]
hmlingesh: Hello h4xnoodle,
in this image https://github.com/cloudfoundry/docs-bosh/blob/nats-rotation/content/images/nats_rotation.png the last section is representation seems to be incorrect. I am referring to "Removing CA1 from all vms" . Please confirm it. Thanks !!!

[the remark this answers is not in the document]
h4xnoodle: @hmlingesh I assume you mean that the last step shows CA1 for the other components like the director still. You're correct, all components should be as the previous step, with the addition of VMs having the new CA. Thanks for pointing it out

content/nats-ca-rotation.md
```diff
@@ -0,0 +1,228 @@
...
+    certificate: ((nats_server_tls_2.certificate))
+    private_key: ((nats_server_tls_2.private_key))
+
+- type: replace
+  path: /instance_groups/name=bosh/properties/nats/tls/client_ca?
+  value:
+    certificate: ((nats_ca_2.certificate))
+    private_key: ((nats_ca_2.private_key))
+
+- type: replace
```
ryanmoran: Am I missing something or aren't the certificates that we replace below already replaced in `add-new-ca.yml`?
- `nats/tls/client_ca`
- `nats/tls/director`
- `nats/tls/health_monitor`

pivotal-dale-wick: Yes, they were regenerated in the `add-new-ca.yml` step.  It is just more orderly to still have them replaced in this step, in my opinion.

PR #541, opened 2018-07-23
Fixed markup

yogendra:
In "Uploading to the Directory", second bullet point, code block did not have proper indentation


dpb587-pivotal:
Thanks!

PR #547, opened 2018-07-31
correct Content-Type header for POST /configs

cwlbraa:
`text/json` results in a 404.


dpb587-pivotal:
Thanks!

Also... a little unfortunate it causes a 404 status.

voelzmo:
@dpb587-pivotal about the HTTP 404: that's sinatra, right? Or can we change something about this?

dpb587-pivotal:
@voelzmo I'm not sure; probably? I'd traditionally expect to see a 415 or 400 in this case.

PR #552, opened 2018-08-09
Add documentation about cert authentication

medvedzver:


dpb587-pivotal:
Waiting to merge until the next release of vsphere. We should then add a note to those properties to document the version at which they were introduced.

PR #556, opened 2018-08-15
Add guide for rotating TLS-configured blobstore CA

h4xnoodle:
Includes versions for stemcells/director which support TLS blobstore.

This PR https://github.com/cloudfoundry/docs-bosh/pull/540 should also adjust the placement of its doc in the menu as shown in this PR.

https://www.pivotaltracker.com/story/show/158608191


content/blobstore-tls.md
```diff
@@ -0,0 +1,83 @@
...
+        bosh:
+          blobstores:
+          - provider: dav
+            options:
+              endpoint: https://((internal_ip)):25250
+              user: agent
+              password: ((blobstore_agent_password))
+              tls:
+                cert:
+                  cat: ((blobstore_ca.certificate))
```
dpb587-pivotal: 🐱

h4xnoodle: Good catch!

hmlingesh:
Hello h4xnoodle, 
in this image https://github.com/cloudfoundry/docs-bosh/blob/nats-rotation/content/images/nats_rotation.png the last section is representation seems to be incorrect.  I am referring to "Removing CA1 from all vms" . Please confirm it.  Thanks !!!

content/blobstore-tls.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Ignored or Stopped Instances
+
+Take note of any ignored instances with `bosh instances --details`. These instances will not be considered when redeploying or recreating instances. Ignored instances will only receive new certificates when recreated.
+
+Any instances stopped with `bosh stop` will not receive the new CA or certificate until recreated. Performing `bosh start` on these instances will transition the VM to running successfully. However, if a redeploy does not trigger a recreate, this instance will fail to update. The agent consults the blobstore for templates during VM lifecycle changes.
+
+Instances that have been stopped with `bosh stop --hard` have been destroyed, keeping the persistent disk if applicable. Performing a `bosh start` on those instances will recreate the VM. These instances will then receive the new CA.
+
+## Introducing a New Certificate Authority (CA)
```
pivotal-jamil-shamy: @h4xnoodle the steps specified can introduce a situation where deployments will fail (as you have mentioned in the "Troubleshooting" section).

Let's say an operator performs the steps here, and then before attempting a recreate of all the VMs, a critical deployment need to happen.
This deployment will fail unless a VM recreation is performed, which can be very costly from time perspective, especially in production environments.
I think we should have a list of steps for blobstore cert rotation that is more immune to these situations and more stable; where we probably need to concatenate the old and new CAs together.

wdyt ?

pivotal-jamil-shamy:
@hmlingesh probably you wanted to comment on the NATS rotation PR, not this PR ?

hmlingesh:
Thanks, @pivotal-jamil-shamy  I have commented on PR #540.

content/blobstore-tls.md
```diff
@@ -0,0 +1,83 @@
+!!! note
+    Applicable for director version 266.2.0+
+
+    Linux stemcells v3586.5+ (agent 2.83.0)
```
pivotal-jamil-shamy: For reference, wondering what we relied on to specify this stemcell version ?

content/blobstore-tls.md
```diff
@@ -0,0 +1,83 @@
...
+
+    Windows 2016 stemcell v1709.9+, Windows 2012R2 v1200.20+ (agent 2.110.0)
```
pivotal-jamil-shamy: also for windows, wondering what we relied on to specify this stemcell version ?

h4xnoodle: https://github.com/cloudfoundry/bosh-agent/commit/55990481ff13b5991bcec77f6e707809b2aa14f9 this commit & consulting with the windows bosh team.

content/blobstore-tls.md
h4xnoodle: For linux stemcells I looked at the agent version that first had the ability to use the blobstore over TLS which was 2.83.0: 
- bosh-davcli took on the modern config in https://github.com/cloudfoundry/bosh-davcli/commit/aee928c4e77e19901f525b6e75c7307eaf52accd starting in 0.0.26
- bosh-davcli 0.0.28 (0.0.26 included) was included in agent starting at 2.83.0 in a bump. https://github.com/cloudfoundry/bosh-agent/commit/c0812113d7ecf9c0df8555b346cef15bf2a1e435

content/blobstore-tls.md
h4xnoodle: My first thoughts are who is the audience of this documentation and if we are to describe every single scenario that someone could have vs the 'happy path' guide. I think we could add a note, and describe what a work-around is in the troubleshooting section. 

Ultimately I think we should be able to eventually rotate without all these steps if we can send the new CAs/process them over NATS.

pivotal-dale-wick:
The link added in the doc for NATS CA rotation might not work until the NATS pull request is accepted and merge to master.

content/blobstore-tls.md
dsboulder: +1 to what Jamil said.  You can't download logs, run colocated errands, or update code on VMs with blobstore CAs that don't match the director.

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+ -o ~/workspace/bosh-deployment/misc/blobstors-tls.yml \
+ -o add-new-blobstore-ca.yml \
+ -o ... additional opsfiles \
+ --vars-store ./creds.yml \
+ -v ... additional vars
+```
+
+* Adds new variables for the new CA/certificates/private_key.
+* The director is given a modified CA with the original CA and the new CA concatenated as `((blobstore_server_tls.ca))((blobstore_server_tls_2.ca))`.
+* The director is updated to use certificates and keys generated by the new CA.
```
pivotal-jamil-shamy: > The director is updated to use certificates and keys generated by the new CA.

There is no mTLS here. The director will only get client certs here.

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+    type: certificate
+    options:
+      ca: blobstore_ca_2
+      common_name: ((internal_ip))
+      alternative_names: [((internal_ip))]
+```
+
+### Step 2: Recreate all VMs, for each deployment. {: #step-2}
+
+VMs need to be recreated in order to receive new certificates generated from the new Blobstore CA being rotated in. If not recreated, agents will continue to attempt communication with blobstore  verifying against the old certificates CA, which will not be verifiable against the new CA.
```
pivotal-jamil-shamy: > VMs need to be recreated in order to receive new certificates generated from the new Blobstore CA being rotated in.

VMs will only get the CA, no mTLS here

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+ -o remove-old-blobstore-ca.yml \
+ -o ... additional opsfiles \
+ --vars-store ./creds.yml \
+ -v ... additional vars
+```
+
+`remove-old-blobstore-ca.yml`
+
+* `blobstore.tls.ca` is updated to remove the old CA from the concatenated CAs.
+* The director will only new certificates. The CA is updated to be just the new CA.
```
pivotal-jamil-shamy: > The director will only new certificates. The CA is updated to be just the new CA.

Looks like mTLS again here

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+* The blobstore server is updated to use a new certificate and private key, generated by the new CA.
+* All components can communicate using the new CA.
+* Note that the bosh-lite.md and bosh-lite-docker.md ops files if used will need to be adapted for the variable generation.
```
pivotal-jamil-shamy: > Note that the bosh-lite.md and bosh-lite-docker.md ops files if used will need to be adapted for the variable generation.

Not sure what do you mean here ?

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+instance_groups:
+- name: bosh
+  properties:
+    agent:
+      env:
+        bosh:
+          blobstores:
+          - provider: dav
+            options:
+              endpoint: https://((internal_ip)):25250
```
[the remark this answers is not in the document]
h4xnoodle: This was straight from bosh-deployment, so we'll leave it like this, as an example.

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Ignored or Stopped Instances
+
+Take note of any ignored instances with `bosh instances --details`. These instances will not be considered when redeploying or recreating instances. Ignored instances will only receive new certificates when recreated.
+
+Any instances stopped with `bosh stop` will not receive the new CA or certificate until recreated. Performing `bosh start` on these instances will transition the VM to running successfully. However, if a redeploy does not trigger a recreate, this instance will fail to update. The agent consults the blobstore for templates during VM lifecycle changes.
+
+Instances that have been stopped with `bosh stop --hard` have been destroyed, keeping the persistent disk if applicable. Performing a `bosh start` on those instances will recreate the VM. These instances will then receive the new CA.
+
+## Introducing a New Certificate Authority (CA) **(for smaller environments)**
```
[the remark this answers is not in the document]
h4xnoodle: We think this simplified strategy still has merit to keep. It will allow operators to work with less downtime, as it requires half of the recreates. This may still be advantageous for users.

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,224 @@
...
+  - just stop and start VMs
+```
+$ bosh -d deployment-name stop
+$ bosh -d deployment-name start
+```
+
+
+## Troubleshooting
+
+* Performing a redeploy across one or more instances that continue to use the old CA will cause the deployment to fail using the smaller deployment technique.
```
h4xnoodle: Remove reference to smaller deployment technique

content/blobstore-ca-rotation.md
```diff
@@ -0,0 +1,224 @@
...
+                    Getting blob from inner blobstore: Shelling out to bosh-blobstore-dav cli:
+                    Running command: 'bosh-blobstore-dav -c /var/vcap/bosh/etc/blobstore-dav.json get d1bccd47-95ad-4516-49bc-0cf42a2782c3 /var/vcap/data/tmp/bosh-blobstore-externalBlobstore-Get731225442',
+                    stdout: 'Error running app - Getting dav blob d1bccd47-95ad-4516-49bc-0cf42a2782c3:
+                    Get https://10.0.1.6:25250/d1/d1bccd47-95ad-4516-49bc-0cf42a2782c3:
+                    x509: certificate signed by unknown authority (possibly because of
+                    "crypto/rsa: verification error" while trying to verify
+                    candidate authority certificate "default.blobstore-ca.bosh-internal")', stderr: '': exit status 1
+```
+
+Any instances that have not been recreated with `bosh recreate` or a redeploy which causes a recreate will cause the above error. Perform a `bosh recreate` on any instances impacting a redeploy.
```
h4xnoodle: add --fix here, as bosh recreate without fix will attempt to drain which won't be possible

PR #560, opened 2018-08-17
Add docs for links API

h4xnoodle:
Describes usage of the links API.

https://www.pivotaltracker.com/story/show/158608432

### DIFF
--- a/content/links-api.md
+++ b/content/links-api.md
@@ -139,7 +139,7 @@ $ uaac curl https://10.0.1.6:25555/links?deployment=zookeeper
 Create an external link with a user-defined consumer. Uses an existing provider.
 
 !!! note
-    The UAA client creating the link must have the `create_link` permission.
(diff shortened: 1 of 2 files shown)


content/links-api.md
```diff
@@ -0,0 +1,225 @@
...
+  }
+]
+```
+
+### `POST /links`: Create Link
+
+Create an external link with a user-defined consumer. Uses an existing provider.
+
+!!! note
+    The UAA client creating the link must have the `create_link` permission.
```
pivotal-jamil-shamy: Looking at the code, `create_link` and `delete_link` are only an internal representation used by the director code. From a UAA client perspective, it looks like the client should have either an bosh admin scope or a bosh team admin authorities.

https://github.com/cloudfoundry/bosh/blob/6f5263dcaf0f4f09923b96cadf60a965b843a7da/src/bosh-director/lib/bosh/director/permission_authorizer.rb#L79-L80

PR #562, opened 2018-08-20
Cross reference credhub documentation

gberche-orange:
Cross reference credhub documentation for additional details over credhub supported credentials types and their associated generation options


mfine30:
Thanks @gberche-orange!

PR #564, opened 2018-08-22
Correct vSphere property: host -> address

dpb587-pivotal:
https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/issues/136

### DIFF
--- a/content/cpi-config.md
+++ b/content/cpi-config.md
@@ -69,7 +69,7 @@ cpis:
 - name: ((vcenter_identifier))
   type: vsphere
   properties:
-    host: ((vcenter_ip))
+    address: ((vcenter_ip))
     user: ((vcenter_user))
     password: ((vcenter_password))
     datacenters:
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
@nehagjain15 mind verifying this is the correct change? I'm pretty confident in it... aside from the fact that it's apparently been like this for a while and we haven't noticed...

nehagjain15:
@dpb587-pivotal cpi expects it to be actually host and not address

https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/blob/00d220c80dba615d33681ba4c84a9f37e3eb3bba/src/vsphere_cpi/lib/cloud/vsphere/config.rb#L222

I think CPI is called with host key instead of address, so transformation might be happening in director

dpb587-pivotal:
Ah, interesting; I missed the "rename" in https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/blob/master/jobs/vsphere_cpi/templates/cpi.json.erb#L9. That's confusing; maybe we can fix that later. I'll try to clarify something in the docs.

Issue #528, opened 2018-06-19
post-stop.md is missing


pc-jedi:
In [content/job-lifecycle.md](https://github.com/cloudfoundry/docs-bosh/blob/master/content/job-lifecycle.md) is a link to `post-stop.md` which does not exist.

SaphMB:
Available now - https://bosh.io/docs/post-stop/

PR #565, opened 2018-08-23
Provide documentation for post-stop script

SaphMB:
As outlined in #528, [Update Lifecycle](https://bosh.io/docs/job-lifecycle/) links to documentation for `post-stop` scripts but this page was non-existent. The current change outlines what `post-stop` scripts are alongside an example of how they can be used.

This PR also fixes some links to existing documentation that could not be viewed as they were not reflected in `mkdocs.yml`.


belinda-liu:
Thank you for the PR @SaphMB!

PR #544, opened 2018-07-27
Document ssh+socks5 scheme of BOSH_ALL_PROXY

jamesjoshuahill:

PR #566, opened 2018-08-23
Fix example BOSH_ALL_PROXY variable

jamesjoshuahill:
Sorry, we got it wrong in https://github.com/cloudfoundry/docs-bosh/pull/544.


dpb587-pivotal:
Oops_thanks

PR #572, opened 2018-09-12
Fix the result of the 'info' CPI method

bingosummer:
The `info` CPI method should return a hash. `stemcell_formats` is one key of the hash.
See: https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/blob/v72/src/bosh_aws_cpi/lib/cloud/aws/cloud.rb#L482-L486


pivotal-mp:
We don't see a `cpi_information` key at the top level, so it might be slightly confusing. Would you mind removing the named key there and just indicate it's a hash?

bingosummer:
Other methods also give the result a name. For example, https://github.com/cloudfoundry/docs-bosh/blob/master/content/cpi-api-v1-method/has-disk.md. It returns `exists`. I followed them to give the result of `info` a name.
I'm fine to remove it if it's confusing.

degaurab:
Hi @bingosummer 
Thanks for the pull request. Merging to master.

PR #575, opened 2018-09-19
Update to align with Softlayer CPI NG

gu-bin:
Update init-softlayer.md and softlayer-cpi.md
Add softlayer-cpi-migration.md


content/init-softlayer.md
```diff
@@ -34,15 +34,29 @@ VLANs provide the ability to partition devices and subnets on the network. To or
 ---
 ## Step 2: Deploy {: #deploy }
 
-1. Install [CLI v2](cli-v2.md).
+1. Install [BOSH CLI v2](cli-v2.md), make sure to download SoftLayer specific version of bosh-cli [linux](https://bosh-softlayer-artifacts.s3.amazonaws.com/bosh-cli-5.0.1.1-softlayer-linux-amd64) or [mac](https://bosh-softlayer-artifacts.s3.amazonaws.com/bosh-cli-5.0.1.1-softlayer-darwin-amd64).
```
dpb587-pivotal: I don't think we should be encouraging separate CLIs. For general OSS users, it sounded like they can use persistent IPs within IBM Cloud and have the same workflow as all other IaaSes. Was that the only reason users needed this custom `bosh-cli` binaries?

content/softlayer-cpi-migration.md
```diff
@@ -0,0 +1,161 @@
+For the users using legacy Softlayer CPI who want to move to Softlayer CPI NG, please refer to this doc to do the migration.
```
dpb587-pivotal: I think you'd want to use "SoftLayer" capitalization/branding consistently.

content/init-softlayer.md
gu-bin: As many users may use dynamic IP, we'd like to also support it.

dpb587-pivotal:
Hi - can you remove dynamic IP support and custom CLI references from this document?

Given Portable IPs work and do not require an independently maintained fork and different software from the rest of BOSH ecosystem, PIPs should be the standard solution we recommend for the general OSS community.

If you want to maintain documentation in support of dynamic IPs as a SoftLayer-specific adaptation, please feel free to maintain a doc about it in your internal ecosystem.

gu-bin:
@dpb587-pivotal I've updated the docs as you suggested. Please review again. Thanks.

dpb587-pivotal:
Looks fine - thanks!

PR #577, opened 2018-09-26
CPI v2 contract documentation

degaurab:
This include all new contracts and examples.

cc @pivotal-dale-wick 

[#158804777](https://www.pivotaltracker.com/story/show/158804777)


content/cpi-api-v2.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Implementation
+
+API version 2 of CPIs will differ from version 1 by the following:
+
+ - Director will send cpi api_version based on info response from for all CPI calls
+ - Director will also send stemcell `api_version` for all CPI calls
+ - Director will be expecting V2 responses IF
+   - CPI info call provide max supported version as >=2 AND
+   - Stemell api_version is >=2 AND
```
voelzmo: Is stemcell version >=2 really a requirement for Director to expect a CPI API v2 response? I think the response for CPI API v2 should be clearly defined and always be the same. The director can decide to ignore parts of it, because the stemcell doesn't support it, but that's a different story. Having different method signatures might not be easily feasible or even desired in all languages, right?

content/cpi-api-v2.md
```diff
@@ -0,0 +1,115 @@
...
+ * [info](cpi-api-v2-method/info.md)
+ * VM Management
+    * [create_vm](cpi-api-v2-method/create-vm.md)
+    * [delete_vm](cpi-api-v2-method/delete-vm.md)
+ * Disk Management
+    * [attach_disk](cpi-api-v2-method/attach-disk.md)
+    * [detach_disk](cpi-api-v2-method/detach-disk.md)
+
+
+#### Additional methods in V1:
```
voelzmo: I'm not sure what the semantics of 'additional methods' in an API version are. Are you saying those should are optional? I think API methods either are part of an API version or not.

content/cpi-api-v2.md
```diff
@@ -0,0 +1,115 @@
...
+  cloud_properties: {instance_type: r3.8xlarge}
+```
+
+## Methods
+
+- All V1 contracts must still be supported. See [CPI API V1](cpi-api-v1.md).
+- To differentiate V2 calls the caller needs to pass in `"api_version": 2` in the context
+
+### Reference Table (Based on each component version)
+* Reg./reg. : Registry
```
voelzmo: Can we drop the legend and instead write 'registry' everywhere?

content/cpi-api-v2-method/info.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+## Arguments
+
+No arguments
+
+
+## Result
+
+ * `stemcell_formats` [Array of strings]: Stemcell formats supported by the CPI. Currently used in combination with `create_stemcell` by the Director to determine which CPI to call when uploading a stemcell.
```
voelzmo: Could we add how this relates to the stemcell metadata? Do we already have some documentation explaning when which CPI will be selected?

content/cpi-api-v2.md
freddesbiens: We are working on this PR right now in order to improve the content and structure.

content/cpi-api-v2.md
freddesbiens: We are working on this PR right now in order to improve the content and structure.

content/cpi-api-v2.md
miguelverissimo: Will update.

content/cpi-api-v2.md
miguelverissimo: We have updated the title for that section.

content/cpi-api-v2-method/info.md
miguelverissimo: We have updated the info pages to remove the mention about combinations and things. Instead, we have added that information to the `build-stemcell` page. Which references back to the CPI v2 info page.

content/cpi-api-v2-method/create-vm.md
```diff
@@ -0,0 +1,176 @@
...
+        "cloud_properties": {}
+      }
+    }
+  ],
+  "error": null,
+  "log": ""
+}
+```
+
+### Implementations
```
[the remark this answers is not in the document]
bingosummer: +1

content/cpi-api-v2.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Methods
+
+- All V1 contracts must still be supported. See [CPI API V1](cpi-api-v1.md).
+- To differentiate V2 calls the caller needs to pass in `"api_version": 2` in the header of the request.
+
+### Reference Table (Based on each component version)
+
+| Director | CPI | Stemcell  | Should update registry   | Add agent setting to IaaS `user-metadata`   |
```
bingosummer: Does `user-metadata` mean https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/master/stemcell_builder/stages/bosh_azure_agent_settings/apply.sh#L31 ?

According to https://github.com/cloudfoundry/bosh-notes/blob/master/proposals/cpi-api-v2.md, the disk settings are directly returned by CPI method. This solution is not used in the final implement. Correct? According to this new doc, the agent setting (including disk setting) is added to `user-metadata`, and bosh-agent will read from that.

content/cpi-api-v2.md
```diff
@@ -0,0 +1,78 @@
...
+|----------|-----|-----------|----------------------|---|
+| 1  | 1  | 1  | Update registry  | Don't add agent setting to IaaS  |
+| 1  | 1  | 2  | Update registry  | Don't add agent setting to IaaS  |
+| 1  | 2  | 2  | Update registry  | Don't add agent setting to IaaS  |
+| 2  | 2  | 2  | DON'T add anything to registry   | Yes, agent will read `user-metadata` and not call registry  |
+| 1  | 2  | 1  | Update registry  | Don't add agent setting to IaaS  |
+| 2  | 2  | 1  | Update registry  | Don't add agent setting to IaaS (no agent support)  |
+| 2  | 1  | 1  | Update registry (CPI will by default update registry)  | Don't add agent setting to IaaS |
+| 2  | 1  | 2  | Update registry (CPI will by default update registry)  | Don't add agent setting to IaaS |
```
bingosummer: Does the table mean CPI has to handle all these 8 cases?

content/cpi-api-v2-method/attach-disk.md
```diff
@@ -0,0 +1,98 @@
+# attach_disk
+
+Attaches disk to the VM.
+
+Typically each VM will have one disk attached at a time to store persistent data; however, there are important cases when multiple disks may be attached to a VM. Most common scenario involves persistent data migration from a smaller to a larger disk. Given a VM with a smaller disk attached, the operator decides to increase the disk size for that VM, so new larger disk is created, it is then attached to the VM. The Agent then copies over the data from one disk to another, and smaller disk subsequently is detached and deleted.
```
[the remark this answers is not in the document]
h4xnoodle: I think this is valuable information to leave here. It is probably useful for the CPI authors to understand the expected usage of the CPI as we're exposing about other scenarios where `attach_disk` is used.

content/cpi-api-v2-method/info.md
```diff
@@ -0,0 +1,44 @@
...
+ * `api_version` [int]: maximum supported version by CPI.
```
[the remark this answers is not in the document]
h4xnoodle: I think we should leave it as it is. The entire section of the docs for CPI V2 is 'new' so I think highlighting one small 'new' part would be confusing. If the CPI authors are looking at V2 docs, it's irrelevant to add 'new' to select additions.

content/cpi-api-v2.md
h4xnoodle: Yes. This table represents the conditions the CPI must continue to write to the registry and produce the correct agent settings to be stored in the agent. 

Essentially it can be simplified to "Is the stemcell V1 or the V1 contract is being used? -> use registry" and "when all of director, CPI, and stemcell/agent understand v2, then do not use the registry".

content/cpi-api-v2.md
h4xnoodle: @bingosummer the CPI does return the disk attach point in `attach_disk` for V2. This is then used by the director to send the appropriate call to the agent (`mount_disk`) which includes this returned disk location. The director <> agent communication is replacing the registry here. 

For user-metadata, this table represents what needs to be in the agent settings for bootstrapping. As of V2, any settings you would have previously placed in the registry should be merged with the existing settings already entering the user-metadata. For actions such as attaching, mounting, detaching disks, the director communicates with the agent directly instead of the registry as a middle-man. 

For example, with the registry the user-metadata may look like:
```json
{
  "dns": ...
  "networks": ...
   "registry":
   {"endpoint": ...}
}
```
which upon bootstrapping the agent consults the registry for the remainder of the settings.

Without the registry agent settings would resemble something like:
```json
{
  "dns": ...
  "networks": ...
   "agent_id": ...
   "vm": {"name": ..."},
   "disks": {
      "ephemeral":  "/dev/sda1",
      "persistent": {}
}
```
it's important to note that persistent disks are never in the agent settings upon bootstrapping. The director and agent communicate further to attach/mount persistent disks. The persistent disk settings are written to another file in the agent, and the agent consults this file along with the settings deposited in the `user-metadata`. We'll improve the docs around this process.

content/cpi-api-v2-method/create-vm.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+## Arguments
+
+ * `agent_id` [String]: ID selected by the Director for the VM's agent.
+ * `stemcell_cid` [String]: Cloud ID of the stemcell to use as a base image for new VM.
+ * `cloud_properties` [Hash]: Cloud properties hash specified in the deployment manifest under the VM's resource pool.
+ * `networks` [Hash]: Networks hash that specifies which VM networks must be configured.
+ * `disk_cids` [Array of strings] Array of disk cloud IDs for the disks that the created VM will most _likely_ attach. The disk cloud IDs could be used to optimize VM placement so that disks are located nearby.
+ * `environment` [Hash]: Resource pool's env hash specified in the deployment manifest, including initial properties added by the BOSH director as shown below. Additionally, the director will append the following guaranteed values:
```
NautiluX: Shouldn't we document somewhere that this argument is expected to be passed to the agent as is via the agent settings?

PR #578, opened 2018-09-27
Update documentation for links.

xtreme-andrew-su:
- Add details about the different link scenarios
- Add an explicit explicit linking section
- Add the allowed properties for providers
- Add the allowed properties for consumers
- Add FAQ section
- Add details about `nil` providers
- Add details about `nil` consumers

[#158511327](https://www.pivotaltracker.com/story/show/158511327)

Co-authored-by: Dale Wick <dwick@pivotal.io>

### DIFF
--- a/content/links.md
+++ b/content/links.md
@@ -230,7 +230,8 @@ instance_groups:
 ```
 
 ### Implicit linking {: #implicit }
-The default type of links in a deployment are the ones that are not mentioned in the `consumes` section of a job in the deployment manifest, but only defined in the release specification for the job.  This type of linking happens automatically, and is referred to as *implicit linking*.
+
+Implicit links are only defined in the release specification for the job, and are not mentioned in the `consumes` section of a job in the deployment manifest. This type of linking happens automatically. By default, links are implicit.
 
 Implicit linking is not supported between deployments.
 
@@ -251,7 +252,7 @@ Unmatched consumers will cause the deployment to fail unless the consumer is def
 
 If a link `type` is provided by only one job within a deployment, all release jobs in that deployment that implicitly consume links of that `type` will resolve to that provider.
 
-In the following example, it's unnecessary to explicitly specify that [web job](#web-release-spec) consumes the `primary_db` link of type `db` from the [postgres job](#postgres-release-spec), since the postgres job is the only one that provides a link of type `db`.
+In the following example, it's unnecessary to specify explicitly that [web job](#web-release-spec) consumes the `primary_db` link of type `db` from the [postgres job](#postgres-release-spec), since the postgres job is the only one that provides a link of type `db`.
 
 ```yaml
 instance_groups:
@@ -272,15 +273,22 @@ Common use cases:
 - deployment contains multiple components that are expected to communicate between each other and there is no benefit for the operator to configure these connections explicitly
 
 ### Explicit Linking {: #explicit }
-When there are multiple providers of the same `type` in a deployment, or a link from a different deployment is needed by a consuming job in this deployment, the consumer can specify the source in the deployment manifest.  *Explicit linking* is defined as specifying the consumer in the deployment manifest. The consumer will be matched on both name and type.
+
+*Explicit linking* is defined as specifying the consumer in the deployment manifest. The consumer will be matched on both name and type.
+
+There are two use cases that require the use of *explicit* linking.
+* To distinguish between multiple providers of the same `type` in a deployment.
+* To consume a link provided by a different deployment.
+
 
 #### Consumers
+
 Explicitly defined consumers can have the following optional properties:
 
 * **from** [String]: Overrides the name of the provider to consume. This should match the name defined in the provider's release spec or the name defined by provider's `as` property in the manifest.
-* **deployment** [String]: The name of the deployment to consume from. If the deployment provided does not exist, it will fail with an error. If the name specified is the same as the current deployment, it will be considered non-cross-deployment.
-* **network** [String]: The name of a network used by the provider. See [custom network linking](#custom-network).
-* **ip_addresses** [Boolean]: *Default is false* Boolean for preferring usage of ip addresses instead of DNS names. In cases such as dynamic network, ip addresses are not available and will still return DNS names. See [dns](dns.md#links) for more details.
+* **deployment** [String]: The name of the deployment to consume from. If the deployment provided does not exist, the consumer will fail with an error. The default value for this property is the name of the current deployment, which means that consumers are expected to be in the same deployment as the provider.
+* **network** [String]: Network to be used by the consumer. This must match the name of one of the networks defined by the provider. The default value is the provider's default network. See [custom network linking](#custom-network).
+* **ip_addresses** [Boolean]: Instructs the director to use ip addresses instead of DNS names. This property is ignored in the case of dynamic networks, which always use DNS addresses. Defaults to *false*. See [dns](dns.md#links) for more details.
 
 Optional consumers may be specified as `nil` in the deployment manifest to block consumption of any providers.
 
@@ -295,7 +303,8 @@ instance_groups:
 ```
 
 #### Providers
-Explicitly specified providers in the deployment manifest which can be both explicitly and implicitly consumed, can have the following optional properties:
+
+Explicitly specified providers in the deployment manifest can have the following optional properties:
 
 * **as** [String]: Overrides the name of the provider defined in the release spec. Example:
 ```yaml
@@ -308,9 +317,11 @@ instance_groups:
 ```
 * **shared** [Boolean]: *Default is false* Sets whether this provider is consumable from another deployment. See [cross deployment links](#cross-deployment).
 
+This applies whether the consumer is explicit or implicit.
+
 ##### Blocking a Link Provider {: #blocking-link-provider}
 
-Providers that are specified as `nil` in the deployment manifest will block consumption of this provider.
+Providers that are specified as `nil` in the deployment manifest cannot be consumed. 
 
 Deployment Manifest:
 ```yaml
@@ -321,6 +332,9 @@ instance_group:
     provides:
       conn: nil
 ```
+
+A common use case for this is when a provider is optional and the current deployment will possibly use an alternative provider.
+
 
 ### Self linking {: #self }
 
@@ -420,7 +434,7 @@ Common use cases:
 
 ### Custom Provider Definitions {: #custom-provider-definitions }
 
-Additional link providers (called `custom providers`) can be defined for a job through the deployment manifest (or runtime config). Each custom provider needs a name and a type. The name can not already exist in the release spec.
+Additional link providers (called `custom providers`) can be defined for a job through the deployment manifest or runtime config. Each custom provider needs a name and a type. The name cannot already exist in the release spec.
 Adding a custom provider for a job does not require any changes to the job's release; only deployment manifest changes are needed.
 
 !!! note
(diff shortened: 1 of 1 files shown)


content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
...
+
+## Possible Link Scenarios
+
+The rules for uniqueness are described in the tables below:
+
+Providers can be specified in any way you want as long as it is **not** consumed.
+
+| Provider Name | Provider Type | Consumer Type | Allowed |
+|---------------|---------------|---------------|---------|
+| Same          | Same          | No consumer   | True    |
```
mfine30: Is `Same` an example or mean the same as something else? Similar question about `Different` below?

mfine30:
One other bigger question. I'm inclined to say we should avoid the use of `you`/`your` in instructions. Skimming bosh.io it looks like we generally avoid it.

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
       conn: {as: db_conn}
 ```
 
+### Explicit Linking {: #explicit }
+Explicit linking is defined as specifying your consumer in your deployment manifest. The consumer will be matched on both name and type.
+
+##### Providers
+Explicitly defined providers which can be both explicitly and implicitly consumed, can have the following optional properties:
```
pivotal-jamil-shamy: > Explicitly defined providers....

The providers are only defined in the job spec of the release (except for custom providers definitions). In the deployment manifest they can only be aliased, exposed to other deployments (with shared `flag`), and blocked (by specifying nil).

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+instance_groups:
+- name: my_instance_group
+  jobs:
+  - name: my_provider_job
+    provides:
+      bar: {as: new_name}
+```
+* **shared** [Boolean]: *Default is false* Sets whether this provider is consumable from another deployment. See [cross deployment links](#cross-deployment).
+
+Providers that are specified as `nil` will cause your *explicit* consumer to fail resolution.
```
pivotal-jamil-shamy: > Providers that are specified as `nil` will cause your *explicit* consumer to fail resolution.

Probably this can be reworded to mention that specifying `nil` for the provider in the manifest will block it from ever being consumed (implicitly or explicitly). There is no benefit to mention what will happen to the consumer, it is technically up to the the operator to figure out where the consumer needs to get the link from.

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+  - name: db_job
+    provides:
+      database: nil
+```
+
+##### Consumers
+Explicitly defined consumers can have the following optional properties:
+
+* **from** [String]: Overrides the name of the provider to consume. This should match the name defined in the provider's release spec or the name defined by provider's `as` property in the manifest.
+* **deployment** [String]: The name of the deployment to consume from. If the deployment provided does not exist, it will fail with an error. If the name specified is the same as the current deployment, it will be considered non-cross-deployment.
```
pivotal-jamil-shamy: Worth mentioning that it defaults to current deployment

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+* **network** [String]: The name of a network used by the provider. See [custom network linking](#custom-network).
```
pivotal-jamil-shamy: > The name of a network used by the provider

This feels a bit ambiguous to me since it does not mention how will that affect the links. Maybe a brief statement about it would be more beneficial.
Also we probably need to mention what is the default for that (since this parameter is optional)

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+* **ip_addresses** [Boolean]: *Default is false* Boolean for preferring usage of ip addresses instead of DNS names. In cases such as dynamic network, ip addresses are not available and will still return DNS names. See [dns](dns#links) for more details.
```
pivotal-jamil-shamy: > Boolean for preferring usage of ip addresses instead of DNS names

worth mentioning here what will this affect (i.e. the link **instances** addresses and not the `link.address` )

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+instance_groups:
+- name: app_ig
+  jobs:
+  - name: app
+    consumes:
+      database: nil
+```
+
 ### Implicit linking {: #implicit }
+Implicit linking is defined as not having your consumer in your deployment manifest but only lives in your release specification.
```
pivotal-jamil-shamy: > Implicit linking is defined as not having your consumer in your deployment manifest but only lives in your release specification.

It reads as if we can define the consumer in the deployment manifest, but we don't really do that.
Also we probably needs to change the word `lives` to something else ?

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
 
-If a link type is provided by only one job within a deployment, all release jobs in that deployment that consume links of that type will be implicitly connected to that provider.
+If a link type is provided by only one job within a deployment, all release jobs in that deployment that implicitly consume links of that type will resolve to that provider.
 
-Optional links are also implicitly connected; however, if no provider can be found, they continue to be `nil`.
+Providers that are specified as `nil` will cause your *implicit* consumer to fail resolution. Unless your consumer is defined as `optional` in your release.
```
pivotal-jamil-shamy: I feel the previous statement that was deleted "`Optional links are also implicitly connected; however, if no provider can be found, they continue to be nil.`" is clearer

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+instance_groups:
+- name: app_ig
+  jobs:
+  - name: db_job
+    provides:
+      database: nil
+  - name: app
+```
+
+Optional links that can not be satisfied implicitly will return `nil` in the template rendering.
```
pivotal-jamil-shamy: If an optional link can not be implicitly satisfied, I think it returns an error for not being able to resolve the link , before it renders the templates? isn't  that the case ?

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+
+Implicit linking does *not* happen across deployments.
```
pivotal-jamil-shamy: > Implicit linking does *not* happen across deployments.

Maybe: "Implicit linking is not supported across deployments." ?

content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
       - port
       - url
 ```
+___
+
+## Possible Link Scenarios
+
+The rules for uniqueness are described in the tables below:
+
+Providers can be specified in any way you want as long as it is **not** consumed.
```
pivotal-jamil-shamy: > Providers can be specified in any way you want as long as it is **not** consumed.

I am a bit unsure what does this mean ?

content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
...
+
+| Provider Name | Provider Type | Consumer Type | Allowed |
+|---------------|---------------|---------------|---------|
+| Same          | Same          | No consumer   | True    |
+| Same          | Different     | No consumer   | True    |
+| Different     | Same          | No consumer   | True    |
+| Different     | Different     | No consumer   | True    |
+
+Providers with different names or types can always be **explicitly** consumed without conflicts.
```
pivotal-jamil-shamy: also unclear to me what is the intention of this statement ?

content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
...
+| Same          | Different     | Implicit      | True    |
+| Different     | Same          | Implicit      | False   |
+| Different     | Different     | Implicit      | True    |
+
+___
+
+## Links FAQ
+
+Q: What characters are valid for link names?<br/>
+A: You should try to use unicode characters for naming. However you can not begin with a colon (`:`) since the ruby YAML parser treats it as a symbol.
```
pivotal-jamil-shamy: I would lean towards saying we recommend links names to be ascii letters and numbers only, even if we support wider spectrum. What do you think ?

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
...
+
+##### Consumers
+Explicitly defined consumers can have the following optional properties:
+
+* **from** [String]: Overrides the name of the provider to consume. This should match the name defined in the provider's release spec or the name defined by provider's `as` property in the manifest.
+* **deployment** [String]: The name of the deployment to consume from. If the deployment provided does not exist, it will fail with an error. If the name specified is the same as the current deployment, it will be considered non-cross-deployment.
+* **network** [String]: The name of a network used by the provider. See [custom network linking](#custom-network).
+* **ip_addresses** [Boolean]: *Default is false* Boolean for preferring usage of ip addresses instead of DNS names. In cases such as dynamic network, ip addresses are not available and will still return DNS names. See [dns](dns#links) for more details.
+
+Optional consumers may be defined as `nil` in your deployment manifest to block linking.
```
mfine30: Is "block linking" the phrase to use here? I originally read this as block linking be a new type of linking similar to implicit/explicit. Perhaps this could be rephrased to something like: "Set an optional consumer to `nil` to disable (right word?) that link. See below for an example".

content/links.md
pivotal-jamil-shamy: Maybe changing  `defined` to `specified` ?

content/links.md
pivotal-jamil-shamy: Maybe provide what would be the default network ?

content/links.md
pivotal-jamil-shamy: > Providers that are specified as `nil` will cause your *implicit* consumer to fail resolution

We can have multiple providers that can still satisfy the consumer needs, even if one of the providers is provided as `nil`

content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
...
+Q: What characters are valid for link names?<br/>
+A: You should try to use unicode characters for naming. However you can not begin with a colon (`:`) since the ruby YAML parser treats it as a symbol.
+
+Q: When are cross-deployment links resolved?<br/>
+A: They are only resolved during a deployment of the consumer. The provider is ready for consumption after the successful deploy of the provider deployment.
+
+Q: If a cross-deployment provider is deleted what happens to my consumer?<br/>
+A: Nothing happens to your consumer until you redeploy. Recreating your consumer deployment should be using the existing values provided to it. Although, there is no guarantee that the service will be available.
+
+Q: Are releases the only things that can provide and consume links?<br/>
```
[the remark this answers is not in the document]
xtreme-andrew-su: Will update.

content/links.md
```diff
@@ -408,3 +471,52 @@ instance_groups:
...
+A: No, there are other things that can provide links, such as [manual links](links-manual), and [external links](links-api). There are also [custom link providers](links#custom-provider-definitions) which variables can use to [consume some DNS values](dns#dns-variables-integration).
```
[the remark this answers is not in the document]
xtreme-andrew-su: Will update.

content/links.md
```diff
@@ -227,13 +227,76 @@ instance_groups:
       conn: {as: db_conn}
 ```
 
+### Explicit Linking {: #explicit }
+Explicit linking is defined as specifying your consumer in your deployment manifest. The consumer will be matched on both name and type.
```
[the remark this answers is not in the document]
xtreme-andrew-su: As discussed, the word `defined` is talking about "explicit linking" and specifying was an okay choice of word here.

content/links.md
xtreme-andrew-su: Will update.

content/links.md
xtreme-andrew-su: Will update.

content/links.md
xtreme-andrew-su: This is the case before the refactor and still remains true. The idea is that the release author needs to rely on `if_link` and not just `link`.

content/links.md
xtreme-andrew-su: Will update.

content/links.md
xtreme-andrew-su: Will update to: "Optional consumers may be specified as `nil` in the deployment manifest to block consumption of any providers."

content/links.md
xtreme-andrew-su: Will update and provide examples.

content/links.md
pivotal-dale-wick: Updated with more context and suggestions for resolving issues.

PR #580, opened 2018-10-01
Updates to vSphere CPI documentation

nehagjain15:
Add documentation for Managed_networks and VM Groups.

### DIFF
--- a/content/vsphere-cpi.md
+++ b/content/vsphere-cpi.md
@@ -122,6 +122,7 @@ Schema for `cloud_properties` section:
 * **nested\_hardware\_virtualization** [Boolean, optional]: Exposes hardware assisted virtualization to the VM. Default: `false`.
 * **datastores** [Array, optional]: Allows operator to specify a list of ephemeral datastores, datastore clusters for the VM. Datastore names are exact datastore names and not regex patterns. At least one of these datastores must be accessible from clusters provided in `resource_pools.cloud_properties`/`azs.cloud_properties` or in the global CPI configuration. Available in v23+. Datastore Clusters can be specified as an array of datastore cluster names. Available in v47+
 * **datacenters** [Array, optional]: Used to override the VM placement specified under `azs.cloud_properties`. The format is the same as under [`AZs`](#azs).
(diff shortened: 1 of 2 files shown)


content/vsphere-cpi.md
```diff
@@ -32,6 +32,19 @@ Schema for `cloud_properties` section used by manual network subnet:
 
 * **name** [String, required]: Name of the vSphere network. Example: `VM Network`.
 
+### Managed Networks ###
+Create T1 router and attach it to T0 router. It also creates virtual switch and attaches it to T1 router.
+So every subnet in network will have its own switch. Route advertisement is enabled on T1 router.
```
mfine30: Is this paragraph intended to describe steps that the operator should take or to explain what the Managed Networks feature does? i.e. my recollection is that the T0 router is a prerequisite, but BOSH will create the T1 router.

content/vsphere-cpi.md
```diff
@@ -32,6 +32,19 @@ Schema for `cloud_properties` section used by manual network subnet:
...
+
+Schema for `cloud_properties` section used by managed network subnet:
+
+* **name** [String, required]: Name of the vSphere network. Example: `Subnet 1`.
```
mfine30: Is this a logical name used inside of a deployment manifest; if so should it be something more akin to `subnet_1` or something without spaces/odd characters.

content/vsphere-cpi.md
nehagjain15: This is to explain what managed networks does. Should I instead update it with following description? And you are right T0 router is a prerequisite

"Automates creation of network on NSX-T. New network will have set of T1 Router and virtual switch"

content/vsphere-cpi.md
nehagjain15: Will update it

content/vsphere-cpi.md
mfine30: I would rephrase it to "Creates a T1 router and attaches it to T0 router. It will also create a virtual switch and attach it to the T1 router"

nehagjain15:
@mfine30 I have updated the docs

mfine30:
Merged! Thanks!

PR #593, opened 2018-10-25
Introduce stemcell block under releases

ansd:
This was introduced with [cli v2.0.30](https://github.com/cloudfoundry/bosh-cli/releases/tag/v2.0.30).

Co-authored-by: Max Becker <max.becker@sap.com>


dpb587-pivotal:
👋

PR #599, opened 2018-11-16
add ubuntu-18.04 or higher install 

blue-bird1:
in ubuntu-18.04, `libreadline6 libreadline-dev` been deleted. Should be replaced with `libreadline-dev `


blue-bird1:
wait i CLA  process

dpb587-pivotal:
Hi - thanks for pointing this out and taking the time to sign the CLA. You're technically right, it's now referred to as `libreadline-dev` in bionic, but since `libreadline6-dev` is still a [virtual package](https://packages.ubuntu.com/bionic/libreadline6-dev) and the command should still technically works, I'm inclined to leave the current instructions for the sake of simpler steps for users.

Let us know if you notice any other inaccuracies!

PR #600, opened 2018-11-16
Fix pages list

7hunderbird:
When I was trying to run the docs locally I got an error with the `mkdocs.yml`

```shell
yaml.parser.ParserError: while parsing a block mapping
  in "/docs/mkdocs.yml", line 194, column 7
expected <block end>, but found '-'
  in "/docs/mkdocs.yml", line 195, column 7
```

I ran the `mkdocs.yml` through a yamllint tool to confirm the error was around line 194 and the `RPC Interface`.  When I read the context I saw that the nested keys under there were actually [redundant anyway](https://github.com/cloudfoundry/docs-bosh/commit/5fada5cb34956b79bf8217d05245f65fb634276e#diff-45ee620855dad847bea76b49932ad7db) for the cpi-api-v1 pages.


dpb587-pivotal:
Looks like this has already been fixed in 81ee5fb27b4b3def96aeb5226872c53e290e7ba0 - thanks for pointing it out.

PR #601, opened 2018-11-16
Specific commands

7hunderbird:
When I was setting up the local server I wanted to understand more explicitly the requirements to update the git submodule and then have Docker.

My hope is that this will help future contributors get up and running more efficiently.


deniseyu:
thanks @7hunderbird I've cherry-picked the README change and have merged it.

PR #602, opened 2018-11-19
Ruby is required during CPI installation

polandy:


dpb587-pivotal:
Thanks for the pull request. I changed the link to a page which mentions additional dependencies, like Ruby.

PR #603, opened 2018-11-20
hard/soft affinity, manifest V2

kinjelom:
Information about OpenStack's hard and soft affinity. 
Examples for deployment manifest v1 and v2.


dpb587-pivotal:
@voelzmo I can't speak to accuracy/content of the PR (I leave that to you), but perhaps we can eventually split this off into an OpenStack-specific doc like the Multi-Homed VMs page or something?

PR #606, opened 2018-11-23
Fix cpi urls

freddesbiens:
Replace cloudfoundry-incubator by cloudfoundry for all CPIs that recently graduated from incubation.


dpb587-pivotal:
Woah, that's a lot of commits. For this sort of thing next time, I'd recommend you try the Squash + Merge option (find it with the arrow drop down next to the regular green Merge button). It helps keep the change history sane so it's not overrun with a full page of the same type of changes :)

dpb587-pivotal:
(👍 on the actual change)

PR #608, opened 2018-11-29
Add documentation for the templates on tmpfs feature

xoebus:
Story: https://www.pivotaltracker.com/story/show/161906558

screenshot from 2018-11-29 11-00-30

screenshot from 2018-11-29 11-00-51


content/job-tmpfs.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+```
+- name: bosh
+  ...
+  properties:
+    ...
+    director:
+      ...
+      enable_nats_delivered_templates: true
```
mfine30: Is this a new property or one that's been in the director for a while? 

I'm assuming this manifest is actually part of the director manifest? not the deployment manifest like below?

xoebus: Yep, it's been there for a while. It's a director property.

PR #611, opened 2018-12-03
Documentation for the /director/certificate-expiry endpoint

miguelverissimo:
[#160199794](https://www.pivotaltracker.com/story/show/160199794)

Signed-off-by: Dale Wick <dwick@pivotal.io>

### DIFF
--- a/content/director-api-v1.md
+++ b/content/director-api-v1.md
@@ -92,6 +92,68 @@ curl -s -k https://192.168.50.4:25555/info | jq .
 
 ---
 
+### `GET /director/certificate_expiry`: Director Certificate Expiry {: #director-certificate-expiry }
+
+#### Response body schema
+
+**[root]** [Array]: List of active director certificates.
+
+  - **certificate_path** [String]: Name of the certificate property.
+  - **expiry** [String]: iso8601 format date of the expiry in UTC.
+  - **days_left** [Integer]: number of full days left rounded down (negative value are expired)
+
+#### Notes:
+
+- requires read scope authentication such as bosh.read
+
+#### Example
+
+```shell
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Hi - I'm not sure about what this documentation represents (i.e. rfc, proposal, implemented), but I have several initial concerns about this:

 * "List of active director certificates" - what does active vs inactive mean here? Does active mean unexpired?
 * `certificate_path` is referring to, what looks like, property paths. Properties are job-specific. There is no reference to job here which makes me concerned this is not correctly abstracting BOSH's existing concepts or, at the very least, missing information.
 * `certificate_path` is probably verbose, given this is already a certificate-prefixed endpoint.
 * `days_left`
    * this is a derived value from `expiry`. Typically APIs avoid including duplicate information in favor of encouraging consumers to use canonical values (unless there is a very specific reason otherwise).
    * this is one format a consumer may be interested, but hour or seconds is another. In addition to whether it should be included at all, is there a reason hour or seconds wasn't provided?
    * because a day is a relatively large value, 0 is ambiguous (e.g. does it mean less than 1 day, or does it mean it expired?). I consider this another reason to only provide `expiry`.
 * read scope authentication - does this require director/global read, or team read sufficient?

For my own understanding, who is the target user for this endpoint/information? Is there a corollary CLI command being introduced? Does `bosh env` play into this anywhere? Are multiple commands going to be asking for this information?

Let me know if I can help talk through some of these concerns.

Thanks!

dpb587-pivotal:
cc @xtreme-andrew-su

h4xnoodle:
We're going to be removing this endpoint

PR #612, opened 2018-12-04
Document bosh.env contract between director, CPI, agent

FlorianNachtigall:
Further update agent settings and create_vm API request example.

[#162326706](https://www.pivotaltracker.com/story/show/162326706)

Co-authored-by: David Ansari <david.ansari@sap.com>

### DIFF
--- a/content/cpi-api-v2-method/create-vm.md
+++ b/content/cpi-api-v2-method/create-vm.md
@@ -16,7 +16,7 @@ As of V2 of the CPI API contract, `create_vm` returns an array of the resultant 
  * `cloud_properties` [Hash]: Cloud properties hash specified in the deployment manifest under the VM's resource pool.
  * `networks` [Hash]: Networks hash that specifies which VM networks must be configured.
  * `disk_cids` [Array of strings] Array of disk cloud IDs for the disks that the created VM will most _likely_ attach. The disk cloud IDs could be used to optimize VM placement so that disks are located nearby.
- * `environment` [Hash]: Resource pool's env hash specified in the deployment manifest, including initial properties added by the BOSH director as shown below. It gets passed by the CPI to the agent where it can be found in the user data's `env` hash in `/var/vcap/bosh/settings.json`. Additionally, the director will append the following guaranteed values:
+ * `environment` [Hash]: Resource pool's env hash specified in the deployment manifest, including initial properties added by the BOSH director as shown below. Additionally, the director will append the following guaranteed values:
      * `bosh` [Hash]: A collection of properties used by the BOSH Agent, and optionally the CPI.
          * `group` [String]: A description of the requested VM in the format `<director-name>-<deployment-name>-<job-name>`.
          * `groups` [Array]: A collection of descriptions for the requested VM, combining `director-name`, `deployment-name` and `job-name` in a range of strings separated by a `-`.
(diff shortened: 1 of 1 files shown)


content/cpi-api-v2-method/create-vm.md
```diff
@@ -60,14 +60,52 @@ Most CPIs choose to communicate with the default Agent. Hence, the communication
...
   "blobstore": {
-    "provider": "local",
-    "options": { "blobstore_path": "/var/vcap/micro_bosh/data/cache" }
+    "provider": "",
+    "options": {}
   },
-  "env": {}
+  "env": {
+    "bosh": {
+      "password": "",
```
dpb587-pivotal: I would be a little hesitant on trying to document all of these properties. It might be useful to keep an example like `mbus` just to help reinforce structure of the hash, but I would be a little concerned on 1) trying to keep values in sync with what's documented in [`env`](https://bosh.io/docs/manifest-v2/#instance-groups) and 2) CPI authors seeing these enumerated and relying on the contents for some reason.

content/cpi-api-v2-method/create-vm.md
```diff
@@ -16,7 +16,7 @@ As of V2 of the CPI API contract, `create_vm` returns an array of the resultant
  * `cloud_properties` [Hash]: Cloud properties hash specified in the deployment manifest under the VM's resource pool.
  * `networks` [Hash]: Networks hash that specifies which VM networks must be configured.
  * `disk_cids` [Array of strings] Array of disk cloud IDs for the disks that the created VM will most _likely_ attach. The disk cloud IDs could be used to optimize VM placement so that disks are located nearby.
- * `environment` [Hash]: Resource pool's env hash specified in the deployment manifest, including initial properties added by the BOSH director as shown below. Additionally, the director will append the following guaranteed values:
+ * `environment` [Hash]: Resource pool's env hash specified in the deployment manifest, including initial properties added by the BOSH director as shown below. It gets passed by the CPI to the agent where it can be found in the user data's `env` hash in `/var/vcap/bosh/settings.json`. Additionally, the director will append the following guaranteed values:
```
dpb587-pivotal: Is it worth documenting the `settings.json` piece? Technically that's something the agent controls and could change. Just curious if/how this information is useful to folks.

content/cpi-api-v2-method/create-vm.md
FlorianNachtigall: Regarding 1): I agree that it is not optimal trying to keep those in sync and that the manifest-v2 docs would be the more intuitive place for release authors. But how would you suggest to document the `env.bosh.` properties which are only relevant for `bosh create-env` without a registry then? Since they were removed from the CPI's `cloud_provider` section, there is no documentation for CPI authors anymore.

content/cpi-api-v2-method/create-vm.md
dpb587-pivotal: Oh, I like the idea of updating that agent-api-interactions docs with this sort of information. Also looks like it needs some attention since the v2 stuff. Feel free to add as much detail as you're interested in, otherwise create an issue if you want to delegate more thorough updates :)

content/cpi-api-v2-method/create-vm.md
FlorianNachtigall: I think @voelzmo has already given some thoughts to that

videlov:
As discussed we created a [new PR](https://github.com/cloudfoundry/docs-bosh/pull/615) for documenting the `create_vm` arguments. We updated this PR to include only documentation about the contract between agent, CPI and Director. Also we created a new [issue](https://github.com/cloudfoundry/docs-bosh/issues/616) to update the agent-api-interactions documentation.

@videlov and @beyhan

beyhan:
We created a [PR](https://github.com/cloudfoundry/docs-bosh/pull/635) to update the agent interaction documentation.

PR #624, opened 2019-02-07
Update deploying-step-by-step.md

neilcx:
The original link is irrelevant, should be removed.


jfmyers9:
Hi @neilcx,

Thanks for pointing this out. The link is actually intended as a fun easter egg in the BOSH documentation. We don't intend to remove it at the moment, let us know if you have any further questions.

Best,

@jfmyers9 && @belinda-liu, CF BOSH team

PR #633, opened 2019-03-12
Mention necessary permissions for deploy

watarumohawk:
Without the proper permissions, you might encounter errors during deploy.
This is mentioned on https://github.com/cloudfoundry-community-attic/bosh-bootstrap/issues/329#issuecomment-203629013

### DIFF
--- a/content/init-aws.md
+++ b/content/init-aws.md
@@ -162,6 +162,50 @@ Your AWS credentials consist of an Access Key ID and a Secret Access Key. Follow
 
     See [AWS CPI errors](aws-cpi-errors.md) for list of common errors and resolutions.
 
+    !!! note
+        Make sure you have the following permissions on IAM.
+
+    ```shell
+    {
+        "Version": "2012-10-17",
+        "Statement": [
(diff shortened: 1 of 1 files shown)


voelzmo:
Hey @watarumohawk 

Thanks for the PR! We already have similar information at https://bosh.io/docs/aws-iam-users/#create which is also linked from the page you are editing (at the top, https://bosh.io/docs/init-aws/#credentials).

It seems it wasn't very visible for you that the link would contain this kind of information. Is there something that would have helped you find this easier?

watarumohawk:
Ah, I couldn't find the [sample policy](https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/master/docs/iam-policy.json) which is linked on [IAM Users
](https://bosh.io/docs/aws-iam-users/#create).

Now that I think about this PR, I think this PR is not needed, but I guess we maybe add another example error on [Common Errors](https://bosh.io/docs/aws-cpi-errors/). This is just my 2 cents though.

PR #638, opened 2019-03-25
Add documentation on absolute path variables

miguelverissimo:
- Explain how variables are stored
- Explain how to reference an existing variable via its absolute path
- Alert to the anti-pattern of using absolute path to share variables
between deployments

[#163562315](https://www.pivotaltracker.com/story/show/163562315)

### DIFF
--- a/content/cli-int.md
+++ b/content/cli-int.md
@@ -152,3 +152,28 @@ bosh interpolate base.yml --vars-store=creds.yml
 cat creds.yml
 # admin_password: vbvdhjbzqelnq7cfyw09
 ```
+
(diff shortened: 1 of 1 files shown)


voelzmo:
I'm not sure if we want to say this way of sharing variables is 
>  strongly discouraged and unsupported.

[bosh-dns uses it specifically this way](https://github.com/cloudfoundry/bosh-deployment/blob/master/runtime-configs/dns.yml#L18), so I'd think this is supported and stays like this at least until we change how we're providing variables to bosh-dns? Are there any ideas or plans to do this? If not, I'd like to remove this sentence from the docs again.

PR #640, opened 2019-04-01
Update runtime-config documentation

miguelverissimo:
Releases used in runtime-config are not uploaded automatically.
- Remove the URL information as it was misleading to be analogous with the deployment `releases` block, i.e. releases would be fetched and uploaded;
- Explicitly call-out that the releases used in `runtime-config` need to already exist or be manually uploaded

### DIFF
--- a/content/runtime-config.md
+++ b/content/runtime-config.md
@@ -66,8 +66,6 @@ addons: []
 
 * **name** [String, required]: Name of a release name used by an addon.
 * **version** [String, required]: The version of the release to use. Version *cannot* be `latest`; it must be specified explicitly.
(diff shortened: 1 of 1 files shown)


voelzmo:
URLs in the `releases` section of runtime configs work. It is a CLI feature, which only applies when calling `bosh update-runtime-config`, not `bosh update-config --type=runtime`. That's why the docs say 
> Works with CLI v2
and the example runtime config in bosh-deployment also uses this: https://github.com/cloudfoundry/bosh-deployment/blob/master/runtime-configs/dns.yml#L4

PR #641, opened 2019-04-02
Improve resurrection config docs

s4heid:
[#164654814](https://www.pivotaltracker.com/story/show/164654814)

### DIFF
--- a/content/resurrector.md
+++ b/content/resurrector.md
@@ -111,16 +111,24 @@ If your deployment consists of 1000 VMs, and you use the defaults, the Resurrect
 ---
 ## Enabling the Resurrector with Resurrection Config {: #enable-with-resurrection-config }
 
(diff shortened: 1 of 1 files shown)


content/resurrector.md
```diff
@@ -156,12 +145,79 @@ rules:
...
+
+```yaml
+rules:
+- enabled: false
+```
+
+One liner for turning resurrection off globally:
+
+```sh
+bosh update-config --type resurrection --name default <(echo "{\"rules\": [{\"enabled\": false}]}")
```
voelzmo: Do we want to use the name `default` here? If users have already some configuration uploaded, this will override it, correct? Suggestion: use some other clearly defined name, such as `bosh.global.disable` or similar.

langered: Sounds good.

PR #649, opened 2019-05-03
Add to vSphere Common Errors

EleanorRigby:
Adding another condition that can trigger `field object_set is not optional` type error.


EleanorRigby:
cc: @ggpaue @yeshwantbabar

PR #653, opened 2019-05-26
fix minor issues for cliv2 commands page

fredwangwang:
image

### DIFF
--- a/content/cli-v2.md
+++ b/content/cli-v2.md
@@ -76,33 +76,6 @@ See [Environments](cli-envs.md).
     bosh unalias-env vbox
     ```
 
-#### Environment {: #environment }
-
-- `bosh -e location environment` (Alias: `env`)
-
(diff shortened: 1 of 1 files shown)


fredwangwang:
More changes than originally found, so changed the PR title as well.

changelog:

* fix styling for delete-env
* dix styling for remove-blob
* Combine `#### Environment {: #environment }` with `### Director Environment {: #director-env}`. They are the same info in two places

mrosecrance:
I think this PR may have only have half of your changes. I see the deduplication for Environment but not the rest of it.

fredwangwang:
It is complete. The Newline added before/after the `remove-blob` and `delete-env` fix the styling issue when rendered by mkdocs

image
image

belinda-liu:
Thanks for the fix!

@belinda-liu && @mrosecrance, CF BOSH team

PR #654, opened 2019-06-03
Add information for vSphere human readable names

neilcx:


content/vsphere-human-readable-names.md
```diff
@@ -0,0 +1,34 @@
...
+    user: ((vcenter_user))
+    enable_human_readable_name: false
+```
+
+2. Deploy the Director
+
+
+# More about human readable names
+
+1. Currently vSphere CPI only support names in ASCII characters ( also need to obey to existing naming policies )
```
EleanorRigby: this is not clear `( also need to obey to existing naming policies )`

neilcx: Discussed and removed.

PR #657, opened 2019-06-28
Use a xenial stemcell in bosh-lite guides

beyhan:
bosh-lite guides use the zookeeper deployment as a sample
deployment. Meanwhile requires the zookeeper deployments a xenial stemcell.
That is why the current bosh-lite guides don't work.


beyhan:
@jfmyers9 Do you update the stemcell version in the `bosh-lite` guides automatically? I wasn't sure that is why I opened this pr.

jfmyers9:
We do not. Thanks for submitting this, makes sense to me.

PR #659, opened 2019-07-11
Add documentation for lifecycle placement rule

edwardecook:
The lifecycle placement rule is referenced within the page but not fully documented within the list of placement rules, this just adds that.

As far as I can tell from the commit that added this feature (https://github.com/cloudfoundry/bosh/commit/111df6a1b82b54869a5872db8c58a6314df34624) it looks like v262 is the first release to include the feature but there may be a more definitive way to determine that.

### DIFF
--- a/content/runtime-config.md
+++ b/content/runtime-config.md
@@ -122,6 +122,7 @@ Available rules:
     * **name** [String, required]: Matching job name.
     * **release** [String, required]: Matching release name.
 * **instance_groups** [Array of strings, optional]: Matches based on instance group names. Available in bosh-release v268+.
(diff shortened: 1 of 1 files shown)


content/runtime-config.md
```diff
@@ -122,6 +122,7 @@ Available rules:
     * **name** [String, required]: Matching job name.
     * **release** [String, required]: Matching release name.
 * **instance_groups** [Array of strings, optional]: Matches based on instance group names. Available in bosh-release v268+.
+* **lifecycle** [String, optional]: Matches based on lifecycle type, either `service` or `errand`. Available in bosh-release v262+.
```
mmanciop: I think it might be worth to specify what a `service` is, i.e., anything that is not an `errand` VM, and that `colocated` errands may actually run on `service` VMs.

PR #661, opened 2019-07-16
Update order of bosh env alias commands

niall-byrne:
Sequence was out of order for creating the bosh alias (not authorized to use `alias-env` before exporting client credentials)


h4xnoodle:
Thanks @niall-byrne

PR #666, opened 2019-07-23
Update documentation around converge_variables and update_mode

jfmyers9:
[#167334368](https://www.pivotaltracker.com/story/show/167334368)


voelzmo:
Hey @mfine30 @jfmyers9 sorry for commenting on an already merged thing, but could you also add the default `update_mode`? From the current docs it is a bit unclear when and if I have to specify this.

mfine30:
@voelzmo totally agree. We realized that as well after I merged it so I added the default without going through a PR. Check it out live ;)

voelzmo:
My bad, thanks!

Issue #660, opened 2019-07-12
Wrong information on persistent disk


ywei2017:
We just found out the statement "If you attach a persistent disk to a virtual machine and then stop, terminate, or delete the VM, your persistent disk data remains intact" is inaccurate.

In vSphere at least, if you delete the BOSH Director, the attached persistent will also be deleted. This is the most disastrous since you lose the whole control plane and very likely need to rebuild the whole foundation.

- In the case of BOSH Director, the VM can be shutdown, but NOT deleted, if you want to keep the persistent disk.

beyhan:
Hi @ywei2017,

could you please provide the link to the documentation you are referencing.

ywei2017:
Hi, @beyhan See the 2nd paragraph in [Persistent Disks](https://bosh.io/docs/persistent-disks/):

"If you attach a persistent disk to a virtual machine and then stop, terminate, or delete the VM, your persistent disk data remains intact. Attaching the persistent disk to another VM allows you to access your data."

I did more testing yesterday, and can confirm that "deleting the VM" will definitely wipe out the persistent disk -- for all VMS (not just the director).

Thanks
Yansheng

beyhan:
Hi @ywei2017,

can you provide us with steps to reproduce the issue. Also the versions for director, CPI and stemcell you are using.

For all persistent disks created by the Director and attached to a VM, there is the concept of [orphaned disks](https://bosh.io/docs/persistent-disks/#orphaned-disks). When a VM is deleted by the Director, the persistent disk will stay as it is for the time configure in [max_orphaned_age_in_days](https://bosh.io/jobs/director?source=github.com/cloudfoundry/bosh#p%3ddirector.disks.max_orphaned_age_in_days).

ywei2017:
@beyhan It's true that persistent disks are orphaned, and kept when BOSH Director deletes a deployment. But the documentation uses the general phrase "delete a VM", which is generally understood as from the IaaS console.

We use vSphere. From the vSphere console, right click on a VM, and "Power Off Guest OS", then "Delete from Disk", which is the typical sequence of "delete a VM". When you do that, any persistent disk attached to the VM is gone.

To address the confusion, there needs to be a **WARNING**, like what I put in my original post. This is particularly important concerning the Director itself, and it happened to us last week.

Thanks
Yansheng

beyhan:
[...]

PR #670, opened 2019-07-28
Clarify peristent disk behavior when a VM is deleted

ywei2017:
Provide clarity of persistent disk behavior when a VM is deleted. See [bug 660](https://github.com/cloudfoundry/docs-bosh/issues/660)

### DIFF
--- a/content/persistent-disks.md
+++ b/content/persistent-disks.md
@@ -3,7 +3,11 @@
 
 Instance groups may need to store persistent data.
 
-If you attach a persistent disk to a virtual machine and then stop, terminate, or delete the VM, your persistent disk data remains intact. Attaching the persistent disk to another VM allows you to access your data.
+If you attach a persistent disk to a virtual machine and then stop the VM, your persistent disk data remains intact. Attaching the persistent disk to another VM allows you to access your data.
+
+** WARNING: If you terminate or delete a VM, the fate of the persistent disk depends on the IaaS provider. For example **
+* In AWS, the default behavior is to keep the persistent disk when you delete a VM.
+* But in vSphere, the persistent disk is permanently destroyed if you right click and pick "Delete from Disk" on a VM.
 
 Persistent disks are kept for each instance under the following circumstances:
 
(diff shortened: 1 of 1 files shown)


h4xnoodle:
Thanks for your contribution.

I think we could further clarify that `bosh delete-vm` will not perform the same operation as an IaaS-level 'delete'. From Bosh's perspective, we don't really have control over what an IaaS decides to do for delete. This is out of the scope of Bosh's behaviours. In general, _all_ VM operations _should_ be performed by Bosh, and Bosh can recover well from a few situations that use non-Bosh operations to manipulate VMs.

Could you add a note on how `bosh delete-vm` will perform the expected way as per the original doc, but an IaaS-level delete will have IaaS-specific behaviours?

Thanks,
Rebecca, BOSH

ywei2017:
@h4xnoodle Sure, no problem. See the updated P/R.

Thanks
Yansheng

content/persistent-disks.md
```diff
@@ -3,7 +3,13 @@
 
 Instance groups may need to store persistent data.
 
-If you attach a persistent disk to a virtual machine and then stop, terminate, or delete the VM, your persistent disk data remains intact. Attaching the persistent disk to another VM allows you to access your data.
+If you attach a persistent disk to a virtual machine and then stop the VM (such as `bosh delete-vm`), your
```
h4xnoodle: "stop" and "bosh delete-vm" aren't really the same thing here. `bosh stop` would likely be more what operators perceive as "stopped". 

Perhaps something like
> If you attach a persistent disk to a virtual machine and then delete the VM via `bosh delete-vm`, your persistent disk data remains intact.

h4xnoodle:
@ywei2017 would you mind making the distinction between 'stop' and 'delete' as per the note?

ywei2017:
@h4xnoodle Good idea. I just updated as suggested.

Thanks
Yansheng

h4xnoodle:
Looks good! @mfine30 will end up merging.

PR #675, opened 2019-08-09
Update manifest-v2.md

FelisiaM:
Change `options` to `optional` for the details about configuring `update_mode`


mfine30:
Thanks @FelisiaM . I know you're part of CF;  before I merge, can you make your org membership public and close/re-open this PR so that dreddbot stops complaining?

PR #678, opened 2019-08-14
🐞 Debugging Director DB has correct path

cunnie:
fixes:
```
-bash: /var/vcap/jobs/director/bin/director_ctl: No such file or directory
```

fixes: incorrect output of `RenderedTemplatesArchive.count` (0 not 3). We use a different model, `Deployment`, which is more familiar.

Slight re-wording for more idiomatic English.


h4xnoodle:
Thanks for this!

PR #682, opened 2019-09-25
Azure has updated the tag count restriction

jseely:
Azure has updated the tag count available in Azure Public cloud from 15 up to 50. This does not however effect Azure Stack, Azure National Clouds (GovCloud, Germany, China, etc), or Azure Storage Accounts, though Azure Storage Accounts are expected to support 50 tags in the near future.


ssapra:
Hi @jseely, is any update needed to the Azure tag restrictions since it's been a while since this PR was created.

If you could rebase as well, we can get this merged. Thanks!

PR #687, opened 2019-10-17
Update manifest-v2.md

daniel-keeney:
Adds tips on how to find specific details of your cloud provider, such as vm types, availability zones, networks, and stem cells.


mfine30:
Hi @daniel-keeney, thank you for submitting the PR. This page is more focused on understanding the specific keys and values that are available in the various parts of a deployment manifest rather than how to find find the names for specific AZs in a given environment. 

I think the information you're intending to provide would be better suited if not already present on a page more like https://bosh.io/docs/deployment-basics/ or https://bosh.io/docs/uploading-stemcells/.

PR #694, opened 2019-11-04
Revise UAA example configuration

boergsen:
* add `admin` user to `clients` section (see https://github.com/cloudfoundry/bosh-deployment/blob/85d2f2122bfc2c1f5184c4b10224705fe22d8113/uaa.yml#L40)
* remove outdated property `login` (see https://github.com/cloudfoundry/cf-deployment/issues/85)
* add missing `encryption` block to example

### DIFF
--- a/content/director-users-uaa.md
+++ b/content/director-users-uaa.md
@@ -54,6 +54,12 @@ In this configuration the Director is configured to delegate user management to 
             - bosh.admin
 
         clients:
+          admin:
+            authorities: bosh.admin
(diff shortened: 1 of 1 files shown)


mrosecrance:
Thanks for making the PR, it certainly looks like an update here is long overdue.

The CLA needs to be signed before I can merge it. The trigger to check only happens when a PR is opened or reopened so I'm going to close this, then if you could sign the CLA and reopen it, I can look at it again.

PR #699, opened 2020-01-08
Clarify resurrector behavior for unresponsive agents

xtreme-conor-nosal:
https://www.pivotaltracker.com/story/show/170609374


xtreme-conor-nosal:
scan_and_fix will resurrect unresponsive agents. `bosh cck --auto` will not.

PR #707, opened 2020-01-13
job lifecycle polishing

VasylTretiakov:
* Fixed bullet points
* Aligned note format with the preceding sections of the doc
* Enhanced punctuation


aashah:
There was some weird whitespace, but I just fixed that in a followup commit - f00d8b8

PR #718, opened 2020-02-06
Update jobs.md

mxplusb:
Added advanced reference to inline Ruby methods on known objects.


h4xnoodle:
Thanks for the suggestion! I'm going to close however, as it's been already the case for a while that release authors can manipulate properties as they see fit, within their own release templates. In general, it's "just an erb file" so you can really do whatever you want all the time. :)

PR #722, opened 2020-05-01
Update cli-v2-install.md

proggi64:
The **brew** command line seems to be wrong. The correct folder is "homebrew-tap" instead of only "tap".

When using the original command you will be ask for the GitHub credentials, but an error will occur even after you typed in a valid user and password.
I tried it on my iMac and the corrected command line worked.

### DIFF
--- a/content/cli-v2-install.md
+++ b/content/cli-v2-install.md
@@ -36,7 +36,7 @@ If you are on macOS with [Homebrew](https://brew.sh/), you can install using the
 1. Use `brew` to install `bosh-cli`:
 
     ```shell
-    brew install cloudfoundry/tap/bosh-cli
+    brew install cloudfoundry/homebrew-tap/bosh-cli
     ```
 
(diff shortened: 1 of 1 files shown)


ssapra:
Hi @proggi64 ,

Homebrew's convention is that the `homebrew-` prefix can be left out of the `brew tap` command. See more [here](https://github.com/Homebrew/brew/blob/master/docs/Taps.md#repository-naming-conventions-and-assumptions)

`brew tap cloudfoundry/tap` should make a shallow copy of the Cloudfoundry [homebrew-tap](https://github.com/cloudfoundry/homebrew-tap) at `/usr/local/Homebrew/Library/Taps/cloudfoundry/homebrew-tap`.

While `brew install cloudfoundry/homebrew-tap/bosh-cli` does succeed, the output shows that it converted this command to point at the tap `cloudfoundry/tap`.

Perhaps you can run `brew untap` to remove the repository and try the original command again.

PR #731, opened 2020-09-30
Document node_group cloud_property for az's

rkoster:


rkoster:
Merge when a new google cpi release is cut

Issue #558, opened 2018-08-17
Explain how bosh partitions disks, such that I understand how I can get an ephemeral disk of a certain size


voelzmo:
**Motivation**
As an operator, I want to ensure that my VMs have an ephemeral disk of certain size `x`, say `10GB`. 

There are several factors which influence the size of the ephemeral disk partition mounted at `/var/vcap/data` and bind-mounted at locations like `/tmp` and `/var/log`. I didn't find a central place where this was documented.

**Influencing Factors on ephemeral partition size**
* On which device is the ephemeral partition created?
  * separate ephemeral disk device?
  * root device?
* If separate ephemeral device
  * Calculate swap partition size
    * Do you set a specific value for swap size? If yes, that's your swap partition size
    * If not: How much RAM does your machine have?
      * Is `RAM > (ephemeral device's size/2)`? If yes, `swap partition size = (ephemeral device's size/2)`
      * If  not: `swap partition size = RAM size`
  * `ephemeral partition size = ephemeral device size - swap partition size`
* If no separate ephemeral device, partitions are created on root device
  * same calculations as above, but reduce device size by 3GB, that's your root partition
  * `ephemeral partition size = (root device size -3GB) - swap partition size`
* note: I tried to write the above from memory, so this might be inaccurate. It is just illustrating how much you need to know to set the correct values for disk sizes.

**IaaS differences**
* gcp allows only to set the [root disk size](https://bosh.io/docs/google-cpi/#vm-types)
* openstack allows only to set the [root disk size](https://bosh.io/docs/openstack-cpi/#resource-pools). Alternatively, you could be using a flavor specifying an ephemeral disk device.
* [AWS](https://bosh.io/docs/aws-cpi/#resource-pools) and [Azure](https://bosh.io/docs/azure-cpi/#resource-pools) can be run in both modes, either explicitly using a separate ephemeral disk device, or using the root disk.
* vSphere?

voelzmo:
Bonus content: explain a bit more about the filesystem layout and which kind of data uses up root disk space, ephemeral disk space and persistent disk space. For example: logs seem to be in `/var/vcap/sys/log`, but are actually consuming disk space on the ephemeral disk, because they're symlinked from `/var/vcap/data/sys/log`.

PR #733, opened 2020-11-04
Fix incorrect file paths in content/vm-config.md

Justin-W:
Fixed several file system paths that were previously documented as '/var/vcap/sys/...' to be '/var/vcap/data/sys/...', so that the docs match the file system structure I am actually seeing in VMs when I fly into them.

For example, I see the following files on the actual file system, which seems to conflict with the documented paths (prior to this fix):
- /var/vcap/data/sys/log/dd-agent/agent.log
- /var/vcap/data/sys/log/dd-agent/agent.stdout.log

### DIFF
--- a/content/vm-config.md
+++ b/content/vm-config.md
@@ -9,7 +9,7 @@ BOSH tries to encourage release authors to follow conventions listed below, so i
 
 * `vcap` user: Pre-configured user that comes with the stemcells. Release jobs may run processes under that user. The default password will be random.
 
-* `/etc/logrotate.d/vcap`: Logrotate configuration for `/var/vcap/sys/log/` sub-directories managed by the Agent.
+* `/etc/logrotate.d/vcap`: Logrotate configuration for `/var/vcap/data/sys/log/` sub-directories managed by the Agent.
 
 ---
 ## Release Job and Package Directories {: #jobs-and-packages }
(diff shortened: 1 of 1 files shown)


Justin-W:
Resubmitting PR (as #734) from a different org for CLA purposes.

Justin-W:
Re-opening PR #733, and closing PR #734.

voelzmo:
copying my comment from the other PR to this one:

Hey @Justin-W, thanks for the PR!

The directory `/var/vcap/data/sys` is symlinked to `/var/vcap/sys`, I'm wondering if that was not what you were seeing on your VM. Do you have a specific use-case where it makes a difference if the access to the files is via a symlink or using the direct location?

```
/var/vcap# ls -la
total 40
drwxr-xr-x 10 root root 4096 Nov  5 11:08 .
drwxr-xr-x 12 root root 4096 Sep 28 13:12 ..
drwx------  7 root root 4096 Nov  5 11:08 bosh
drwxr-xr-x 26 root root 4096 Nov  5 11:09 data
drwxr-xr-x  2 root root 4096 Nov  5 11:07 instance
drwxr-x---  2 root vcap 4096 Nov  5 11:08 jobs
drwxr-xr-x  3 root root 4096 Sep 28 13:26 micro_bosh
drwxr-xr-x  5 root root 4096 Nov  5 11:09 monit
drwxr-xr-x  2 root vcap 4096 Nov  5 11:08 packages
drwxr-xr-x  8 root root 4096 Sep  5  2019 store
lrwxrwxrwx  1 root root   18 Nov  5 11:06 sys -> /var/vcap/data/sys
```

Justin-W:
> The directory `/var/vcap/data/sys` is symlinked to `/var/vcap/sys`, I'm wondering if that was not what you were seeing on your VM.

Ah. Then I guess the existing docs are technically correct, but a bit misleading. So I would recommend either adding that symlink equivalency to the docs page and/or accepting my PR.

Since `df -h` output explicitly shows that `/` and `/var/vcap/data/` are separate filesystems with separate mount paths, the current documentation which references `/var/vcap/sys/...` makes it seem that the files under `/var/vcap/sys/` use up the disk space from the `/` drive (which most CPIs don't allow to have a configurable size) instead of the `/var/vcap/data/` drive (which is CPI-configurable).

Also, the existing wording on lines 36 and 38 of the page makes it non-obvious that those lines reference data that is related to the path documented on line 32.

So my issue was not that the runtime behavior was wrong, or that symlinking is breaking anything. It is just that the documentation seemed to conflict with what I was actually observing  within the VM's shell.

voelzmo:
> Since df -h output explicitly shows that / and /var/vcap/data/ are separate filesystems with separate mount paths, the current documentation which references /var/vcap/sys/... makes it seem that the files under /var/vcap/sys/ use up the disk space from the / drive (which most CPIs don't allow to have a configurable size) instead of the /var/vcap/data/ drive (which is CPI-configurable).

I absolutely agree, that we should have a documentation page explaining the filesystem layout and the disk sizes you can influence: #558 

My guess is that the directory always has been `/var/vcap/sys/log` and at some point people decided to put this on the ephemeral disk instead of the root disk (because you can increase the ephemeral disk size easily) and solved this with a symlink without breaking any existing tooling, releases and scripting. This happened some time when we switched from ruby agent to go agent (~2014-2015), so this is just a guess.

I empathize with the confusion you're experiencing when looking at the output of `df -h` and comparing it to the paths where logfiles are referenced in the docs, however, at the same time I feel uncomfortable changing the documentation the the "implementation detail" that at some point the logs were symlinked to be stored on the ephemeral disk. You're right that it is quite a stretch to understand that the log directories discussed in lines 36-38 are part of the ephemeral data mentioned in line 32. Would a `note` that these paths are symlinked from a path on the ephemeral disk be helpful to clear up the confusion you experienced?

Justin-W:
> Would a `note` that these paths are symlinked from a path on the ephemeral disk be helpful to clear up the confusion you experienced?

Yes, that would be sufficient, I think.

And I do agree that there are strong ties between this issue and #558.

PR #734, opened 2020-11-04
Fix incorrect file paths in content/vm-config.md

Justin-W:
Fixed several file system paths that were previously documented as '/var/vcap/sys/...' to be '/var/vcap/data/sys/...', so that the docs match the file system structure I am actually seeing in VMs when I fly into them.

For example, I see the following files on the actual file system, which seems to conflict with the documented paths (prior to this fix):
- /var/vcap/data/sys/log/dd-agent/agent.log
- /var/vcap/data/sys/log/dd-agent/agent.stdout.log

### DIFF
--- a/content/vm-config.md
+++ b/content/vm-config.md
@@ -9,7 +9,7 @@ BOSH tries to encourage release authors to follow conventions listed below, so i
 
 * `vcap` user: Pre-configured user that comes with the stemcells. Release jobs may run processes under that user. The default password will be random.
 
-* `/etc/logrotate.d/vcap`: Logrotate configuration for `/var/vcap/sys/log/` sub-directories managed by the Agent.
+* `/etc/logrotate.d/vcap`: Logrotate configuration for `/var/vcap/data/sys/log/` sub-directories managed by the Agent.
 
 ---
 ## Release Job and Package Directories {: #jobs-and-packages }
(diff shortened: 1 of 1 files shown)


voelzmo:
Hey @Justin-W, thanks for the PR!

The directory `/var/vcap/data/sys` is symlinked to `/var/vcap/sys`, I'm wondering if that was not what you were seeing on your VM. Do you have a specific use-case where it makes a difference if the access to the files is via a symlink or using the direct location?

```
/var/vcap# ls -la
total 40
drwxr-xr-x 10 root root 4096 Nov  5 11:08 .
drwxr-xr-x 12 root root 4096 Sep 28 13:12 ..
drwx------  7 root root 4096 Nov  5 11:08 bosh
drwxr-xr-x 26 root root 4096 Nov  5 11:09 data
drwxr-xr-x  2 root root 4096 Nov  5 11:07 instance
drwxr-x---  2 root vcap 4096 Nov  5 11:08 jobs
drwxr-xr-x  3 root root 4096 Sep 28 13:26 micro_bosh
drwxr-xr-x  5 root root 4096 Nov  5 11:09 monit
drwxr-xr-x  2 root vcap 4096 Nov  5 11:08 packages
drwxr-xr-x  8 root root 4096 Sep  5  2019 store
lrwxrwxrwx  1 root root   18 Nov  5 11:06 sys -> /var/vcap/data/sys
```

PR #742, opened 2021-05-13
Add new nsx-t tagging behavior to vsphere docs

cunnie:
- Add a note to the `tags` cloud config to clarify that it is not the same as tagging in the deployment or runtime config.
- Add a section to the `use_policy_api` setting to indicate that bosh metadata will be added to segment ports after the next release.


ystros:
Thanks @cunnie !

PR #745, opened 2021-06-01
Refine DNS documentation

gberche-orange:
- mention that .bosh TLD is configureable
- mention that /etc/hosts is updated by bosh-agents as specified within https://github.com/cloudfoundry/bosh-agent/blob/b68c7860102f69578f94be6227222f43a6e71635/integration/sync_dns_test.go#L128-L152
- complement DNS query language example with BNF like syntax


jpalermo:
Looks great. Thanks

PR #746, opened 2021-07-22
Adding example for different domains of user and project

andinod:
Due to the recent changes in the bosh-openstack-release and the bosh-release, it's necessary to include an example for the usage of this new feature to connect to Openstack.


rkoster:
Thanks @andinod!

PR #747, opened 2021-07-26
Update vsphere-cpi.md

julian-hj:
As of v64, `datastore_pattern` may be omitted if `datastore_cluster_pattern` is set. This commit makes the docs for `datastore_pattern` and `datastore_cluster_pattern` consistent with `persistent_datastore_pattern` and `persistent_datastore_cluster_pattern` now that the behavior is also consistent.

### DIFF
--- a/content/vsphere-cpi.md
+++ b/content/vsphere-cpi.md
@@ -224,7 +224,7 @@ Schema:
     * **vm_folder** [String, required]: Path to a folder (relative to the datacenter) for storing created VMs. Folder will be automatically created if not found.
     * **template_folder** [String, required]: Path to a folder (relative to the datacenter) for storing uploaded stemcells. Folder will be automatically created if not found.
     * **disk_path** [String, required]: Path to a *disk* folder for storing persistent disks. Folder will be automatically created in the datastore if not found.
-    * **datastore_pattern** [String, required if `datastore_cluster_pattern` is not set]: Pattern for selecting datastores for storing ephemeral disks and stemcells.
+    * **datastore_pattern** [String, required if `datastore_cluster_pattern` is not set or CPI version <= v63]: Pattern for selecting datastores for storing ephemeral disks and stemcells.
     * **datastore\_cluster\_pattern** [String, required if `datastore_pattern` is not set]: Pattern for selecting datastore clusters for storing ephemeral disks. This feature is experimental. Available in v63+.
     * **persistent\_datastore\_pattern** [String, required if `persistent_datastore_cluster_pattern` is not set]: Pattern for selecting datastores for storing persistent disks.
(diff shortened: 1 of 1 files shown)


content/vsphere-cpi.md
```diff
@@ -224,8 +224,8 @@ Schema:
     * **vm_folder** [String, required]: Path to a folder (relative to the datacenter) for storing created VMs. Folder will be automatically created if not found.
     * **template_folder** [String, required]: Path to a folder (relative to the datacenter) for storing uploaded stemcells. Folder will be automatically created if not found.
     * **disk_path** [String, required]: Path to a *disk* folder for storing persistent disks. Folder will be automatically created in the datastore if not found.
-    * **datastore_pattern** [String, required]: Pattern for selecting datastores for storing ephemeral disks and stemcells.
-    * **datastore\_cluster\_pattern** [String]: Pattern for selecting datastore clusters for storing ephemeral disks. This feature is experimental. Available in v63+.
+    * **datastore_pattern** [String, required if `datastore_cluster_pattern` is not set]: Pattern for selecting datastores for storing ephemeral disks and stemcells.
+    * **datastore\_cluster\_pattern** [String, required if `datastore_pattern` is not set]: Pattern for selecting datastore clusters for storing ephemeral disks. This feature is experimental. Available in v63+.
```
klakin-pivotal, line 228: Is this the best way to mention that `datastore_pattern` is optional if `datastore_cluster_pattern` is set, but only on CPI version 64+?

I notice that the docs for `d_c_p` mention that it's available in CPI 63+. If a user was running CPI 63, and they omitted `d_p` and set only `d_c_p`, would they encounter an error, or would the configuration be accepted?

If they would encounter an error, then would you mind updating the PR to make it clear that setting only one of the two is only supported on CPI version 64+?

julian-hj: Hmm, yeah that did occur to me, but I couldn't think of a way to describe that nuance that would actually make things clearer, so I decided that it might be better to just let it go undocumented. Perhaps that was the wrong decision though. I'm just now sure how to put it.

| * **datastore_pattern** [String, required if `datastore_cluster_pattern` is not set or version < v64]: Pattern for selecting datastores for storing ephemeral disks and stemcells.

feels a bit unwieldy to me. WDYT?

klakin-pivotal: I would say

> ...is not set or CPI version <= v63

but otherwise it seems good to me.

(Putting "<=" instead of just "<" makes it less likely that _I'd_ just skim over that detail, and being painfully explicit about what component's version is being talked about is often helpful to me... so hopefully these changes will help others.)

julian-hj: Fair enough. I will make that edit.

PR #748, opened 2021-07-29
Update certificate duration description

pvaramballypivot:
Explains minimum duration logic.
Please wait until a new CredHub release is created that contains the Minimum Certificate Duration logic before merging this PR.

Authored-by: Preethi Varambally <pvarambally@pivotal.io>
Authored-by: Brian Upton <bupton@vmware.com>


klakin-pivotal:
@pvaramballypivot Would you folks mind letting us know when that new version of CredHub is released? That'd help with timing this PR merge correctly.

klakin-pivotal:
Also, if there is a minimum version of CredHub (or the Credhub Bosh Release) required to use the documented feature, does it make sense to mention that version in the documentation? Seems to me like that would reduce user confusion...

pvaramballypivot:
@klakin-pivotal Sorry I missed your comments. Credhub team is reviewing/reconsidering the minimum duration changes. So we might have to wait on this. 
Tagging @peterhaochen47 and @bruce-ricard from the Credhub team. 

cc @ystros @pabloarodas

pabloarodas:
The feature is still in progress so we still need this PR, the CredHub team is already working on the final details to get the new version so we will update the version as soon as it's ready.

PR #749, opened 2021-09-01
Update docs for global config of disk paths

julian-hj:
This change fixes https://github.com/cloudfoundry/bosh-vsphere-cpi-release/issues/210

We incorrectly documented three properties as required that actually have default values. This PR corrects the documentation to reflect the behavior of the product.


rkoster:
Thanks @julian-hj

PR #750, opened 2021-09-01
Clarify that `vcenter_ip` can be a hostname

julian-hj:
- This fixes https://github.com/cloudfoundry/bosh-vsphere-cpi-release/issues/183
- Using a hostname or FQDN sometimes has advantages and we'd like to make it clearer that it is possible in this context.


rkoster:
Thanks @julian-hj, LGTM

PR #751, opened 2021-09-07
Fix indentation issue

dlresende:
The tool currently used to convert MD docs into HTML seems not to be rendering nested items from unordered lists as intended. Maybe because tabs VS spaces or number of spaces required?!

I'm proposing we merge item + nested item to address this.


rkoster:
@dlresende Thanks! Makes sense

PR #752, opened 2021-09-07
Mention bosh-dns-aliases-release

dlresende:
Currently, the docs do not mention the ability to leverage the `bosh-dns-aliases-release` if the Operator wants to add a DNS alias. This is very handy for those willing to keep changes at the manifest level, but cannot change the `bosh-dns` job directly.


rkoster:
@dlresende Thanks! LGTM

PR #753, opened 2021-09-07
Document BPM hardcoded timeouts and standard troubleshooting procedure

bgandon:
Hi,
In this PR, I started documenting the hardcoded "stop" procedure in BPM.
Then I detailed the concept of "PID check" in an adjacent section, and finally added some troubleshooting default procedure, and tips for using the tools provided by default on stemcells.
Cheers


lnguyen:
Thanks for commit!

PR #763, opened 2022-02-03
Docs reflect primacy of CPI v2

aramprice:
- CPI v2 reference has entries for all CPI methods
- CPI v1 reference has deprecation warning, only lists v1 specific methods

See related discussion https://github.com/cloudfoundry/bosh/discussions/2355


lnguyen (approved):
LGTM, thanks @aramprice

ramonskie (approved):
LGTM

PR #766, opened 2022-03-10
Remove unneccessary vars generation

schmidtsv:
In the second ops file, the /variables do not need to be touched. The only 2 cases where they could be not there would be:
- You run Step3 before step 1
- You forget to remove the ops after step 5

This way in both cases the bosh deploy errors and will not cause all VMs to lose connection to BOSH. We internally use it like this for quite a while now and it prevented some incidents already so I think everyone should have that.

### DIFF
--- a/content/nats-ca-rotation.md
+++ b/content/nats-ca-rotation.md
@@ -156,49 +156,6 @@ bosh create-env ~/workspace/bosh-deployment/bosh.yml \
     certificate: ((nats_clients_health_monitor_tls_2.certificate))
     private_key: ((nats_clients_health_monitor_tls_2.private_key))
 
-- type: replace
-  path: /variables/-
-  value:
-    name: nats_ca_2
-    type: certificate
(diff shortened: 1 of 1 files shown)


schmidtsv:
@jpalermo The only thing i remove is that the ops file will not generate the var here. If you don't persist the new nats CA then you are doomed anyways. At the point you run the second ops file then you should already have the nats_2 certs and CA in your bosh secret file. 

I kept the nats_2 in the ops for the filling out of templates. It loads fine from the bosh-secrets.yml unless its missing, which only happens in the 2 error cases I mentioned in the first message.

jpalermo:
Oh yeah, my brain totally forgot about the vars file. 👍

jpalermo (approved):

rkoster (approved):
LGTM

rkoster:
Thanks! @schmidtsv

PR #769, opened 2022-04-19
Improve vendor-package documentation

mvach:
This PR persist the learnings we made while using vendor-package with the previous documentation.


lnguyen (approved):
Thanks!!

ramonskie (approved):
LGTM

PR #770, opened 2022-05-23
Update bosh create-release docs

Benjamintf1:
This commit updates the docs for the bosh create-release command so that they do not imply that the `PATH` positional parameter is an option to create a tgz, but instead it is an option for creating releases from release yml files.

@klakin-pivotal also worked on.


jpalermo:
Looks good thanks!

PR #773, opened 2022-06-23
Component diagram without registry

mvach:
BOSH Registry was removed in v272.0.0.
Here it's removed from the component diagram as well.


rkoster (approved):
Really nice job on the new diagram!

rkoster:
Thanks! @mvach

PR #774, opened 2022-06-28
Add name parameter to cloud configs command documentation

mvach:
This PR depends on: https://github.com/cloudfoundry/bosh-cli/pull/600
This is required to the fix https://github.com/cloudfoundry/bosh-cli/issues/542


friegger (changes requested):
Can you add the information which Director version is required?

jpalermo (approved):
Looks good. Thanks!

friegger (approved):

PR #777, opened 2022-08-27
Document `create` as value of version in manifest

mkocher:
I learned this trick from @acrmp but it usually takes me two tries to remember where create goes and how to specify the local path to the bosh release directory.


rkoster (approved):

aramprice (approved):

rkoster:
Thanks! @mkocher

PR #779, opened 2022-10-12
Document vSphere cpi memory_reservation_locked_to_max

rkoster:
Related PR: https://github.com/cloudfoundry/bosh-vsphere-cpi-release/pull/339


cunnie (approved):
I modified the CPI version from v81 to v82 because v81 has already been published & doesn't include `memory_reservation_locked_to_max`

PR #780, opened 2022-11-18
Add practical examples of bosh handling AZ changes

nouseforaname:
[ #183715145 ]

The behavior of bosh in regards to adding / removing / replacing AZs to a deployment and the subsequent rebalancing of VMs (especially VMs with persistent disks) is somewhat hard to understand. This adds examples for the expected behavior of bosh when attempting to change the AZs for instance_groups in several scenarios.

### DIFF
--- a/content/azs.md
+++ b/content/azs.md
@@ -97,3 +97,384 @@ Task 1442 done
 
 VMs total: 4
 ```
+
+---
+
+## Adding and removing AZs do a deployment.
+
+> **_NOTE:_**
+>   With some CPIs AZS are referenced in Subnet configs of the cloud-config networks block. This means that moving AZs can have networking implications. E.g. networks can be Zonal ( e.g. AWS Subnets are associated to a Zone) or Regional ( e.g. GCP Subnets are associated with a Region). This is important to keep in mind when planning an AZ migration.
+
+
+> **_GENERAL LIMITATIONS:_** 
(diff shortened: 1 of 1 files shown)


content/azs.md
```diff
@@ -97,3 +97,384 @@ Task 1442 done
...
+dummy/3697cb63-5329-4b61-8251-6acd73fe5d8b ... az3 
+dummy/63d450fc-a071-4e19-b0ba-c8fdb147dcce ... az1 
+dummy/528993ea-5e8b-4d7f-8844-98234bcb0575 ... az1 
+```
+
+Since the the above scenario does utilize persistent disks, adding az2 to the list of azs will:
+
+1. Skip redeploying a vm from az1 to az2 to avoid recreating the persistent disk and potentially incur dataloss.
+
+Bosh opts to keep the persistent disk to avoid dataloss. Bosh is not aware about the capabilities, in terms of distributed state, of the software it deploys. Some software architectures have internal features (e.g. Nats is utilizing RAFT) that allow syncinc state within the cluster nodes. Other architectures rely on features provided by their host or additional software (e.g. distributed filesystems) to achieve a similar outcome.
```
randomtask1155, line 215: typo syncinc -> syncing

beyhan (approved):

rkoster (approved):

PR #787, opened 2023-03-20
Documents '--system' and '--all-logs' flags

klakin-pivotal:
This commit:
* Documents the '--system' and '--all-logs' flags for the 'logs' command.
* Documents the "new" 'system' log type. (This is just the contents of `/var/log`... it's not actually new. But we're allowing you to interact with it through the CLI for the first time, so it's a new CLI concept.)
* Notes that downloading 'system' logs requires using Agent v2.516.0 or later. I'm pretty sure that this is the right version.
* Updates the documentation for the logs-filtering '--only' flag. (The examples provided would have generated an empty log tarball. These new examples would actually generate a tarball that has files in it.)

_**Please look over this PR carefully.**_ I'm especially uncertain of the description I added of the "system" log type. I believe that it _should_ be documented, but I'm not at all certain that I've documented it well.

### DIFF
--- a/content/cli-v2.md
+++ b/content/cli-v2.md
@@ -1106,8 +1106,10 @@ See [CPI config](cpi-config.md).
     - `--num=NUMBER` Last number of lines
     - `-q`, `--quiet` Suppresses printing of headers when multiple files are being examined
     - `--job=NAME` Limit to only specific jobs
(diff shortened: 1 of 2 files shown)


selzoc (approved):
LGTM

PR #795, opened 2023-07-10
Document `create-recovery-plan` and `recover` commands

selzoc:
[#185483613, #185483615]

Authored-by: Chris Selzo <cselzo@vmware.com>


content/cli-v2.md
```diff
@@ -1546,6 +1546,26 @@ See [CPI config](cpi-config.md).
 
     See [Manual repair with Cloud Check](cck.md) for details.
 
+#### Create-Recovery-Plan {: #create-recovery-plan }
+
+- `bosh [GLOBAL-CLI-OPTIONS] create-recovery-plan PATH`
+
+    Checks for resource consistency and interactively prompts for problem resolution by instance group.  Optionally allows overriding the current value of `max_in_flight` for each instance group.  This information is then written to a YAML file and is used by the `bosh recover` command.
```
mvach, line 1553: How can I override the `max_in_flight` value  while generating the plan. I don't see any option for that.
Is this done manually in the generated plan afterwards? Then I would document that in the `recover` command only.

selzoc: I have added some examples - hope this clears it up!

content/recover.md
```diff
@@ -0,0 +1,252 @@
...
+    Requires director v277.4.0 and CLI v7.3.0
+
+BOSH provides the `create-recovery-plan` and `recover` CLI commands to repair IaaS resources used by a specific deployment. The underlying machinery is very similar to [Cloud Check](cck.md), with several exceptions:
+
+1. There is a 2-step process: `create-recovery-plan` scans the deployment for problems and then prompts the user to generate a recovery plan, which is saved to a file.  The `recover` consumes that file.
+1. Resolutions to problems are selected by instance group and problem type. The `cloud-check` command asks for a resolution for each particular problem.
+1. When generating a recovery plan, `max_in_flight` can be overriden per instance group.  This can be handy to speed deployment recovery.  The `cloud-check` command uses the `max_in_flight` values in the deployment manfiest.
+
+
+Otherwise, the types of problems and the mechanism by which they are repaired are the same as in the `cloud-check` command.
```
klakin-pivotal, line 11: Might be nice to add a link to <https://bosh.io/docs/cck/#problems>, as the intent seems to be to include that part of the documentation by reference.

I know we did link to the main page, but repeating the information will help some folks who are skimming through the documentation very quickly.

content/recover.md
```diff
@@ -0,0 +1,252 @@
...
+
+1: Skip for now
+2: Reboot VM
+3: Recreate VM without waiting for processes to start
+4: Recreate VM and wait for processes to start
+5: Delete VM
+6: Delete VM reference (forceful; may need to manually delete VM from the Cloud to avoid IP conflicts)
+unresponsive_agent (1): 2
+
+Override current max_in_flight value of '50%'? [yN]:
```
klakin-pivotal, line 70: Might be nice to put either `n` or `<enter>` here to give the reader more of a clue as to what you did here. (Like, if I think about it for a minute, I can determine that you pressed enter, but it'd be nice to SEE that.)

content/recover.md
klakin-pivotal, line 11: Might be nice to add a link to <https://bosh.io/docs/cck/#problems>, as the intent seems to be to include that part of the documentation by reference.

I know we did link to the main page, but repeating the information will help some folks who are skimming through the documentation very quickly.

content/recover.md
```diff
@@ -0,0 +1,252 @@
...
+Succeeded
+```
+
+### Recovery plan {: #recovery-plan }
+
+Each recovery plan has the following schema:
+
+**instance_groups_plan** [Array, required]: The name of instance groups in the deployment to recover.
+
+* **max_in_flight_override** [String, optional]: The `max_in_flight` value to use for problem resolution in the given instance group
```
klakin-pivotal, line 152: Would be good to specify here what valid values for this string are.

Can I put "all"? "Two"? etc, etc, etc. (As a user of this feature, if I didn't look at the example, I wouldn't know that "100%" is a valid value. What else is valid? Fractional percents? Fractions? (so on and so forth))

klakin-pivotal (approved):

selzoc:
@mvach can you finish your review?  I'd like to get these docs merged as the CLI and Director have been released with these changes.  Thanks!

content/cli-v2.md
mvach: Ah, got it.

PR #800, opened 2023-08-10
Document the `--force-latest-variables` flag on `bosh deploy`

selzoc:


content/cli-v2.md
```diff
@@ -960,6 +960,7 @@ See [CPI config](cpi-config.md).
     - `--canaries=NUMBER or PERCENTAGE` Override manifest values for canaries
     - `--max-in-flight=NUMBER or PERCENTAGE` Override manifest values for max_in_flight
     - `--dry-run` Renders job templates without altering a deployment. It will save some state in the director database (like ip reservations) which will be reused on the next deploy
+    - `--force-latest-variables` Causes the director to retreive the latest value of all variables from the config server, overriding their update strategies
```
mvach, line 963: Just to avoid that @rkoster will ask for it :-) 
We should add the minimal director version that supports that option.

PR #801, opened 2023-08-14
Consistenly link the documentation for the version of monit used with BOSH

silvestre:
Use a snapshot archived by the Internet Archive's Wayback Machine as the
manual/documentation for Monit 5.2.5 can no longer be found on the
current website.

### DIFF
--- a/content/create-release.md
+++ b/content/create-release.md
@@ -209,7 +209,7 @@ On a deployed release, a BOSH Agent runs on each job VM.
 BOSH communicates with the Agent, which in turn executes commands in the
 control script.
 The Agent does this using open source process monitoring software called
-[Monit](https://web.archive.org/web/20110816041503/https://mmonit.com/monit/documentation/monit.html).
+[Monit, version 5.2.5](https://web.archive.org/web/20110816041503/https://mmonit.co+m/monit/documentation/monit.html).
 
 The `monit` file for the `web_ui` job looks like this:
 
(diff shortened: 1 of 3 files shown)


aramprice:
This looks good, thank you.

One thought is that it may make sense to change the link text to indicate the monit version and/or indicate that the link is going somewhere other than `mmonit.com`.

Something like `Archived Monit Documentation` or `Monit 5.2.5`, instead of just `Monit` as the link text.

silvestre:
> One thought is that it may make sense to change the link text to indicate the monit version and/or indicate that the link is going somewhere other than `mmonit.com`.
> 
> Something like `Archived Monit Documentation` or `Monit 5.2.5`, instead of just `Monit` as the link text.

Good idea, thanks, done!

content/create-release.md
```diff
@@ -209,7 +209,7 @@ On a deployed release, a BOSH Agent runs on each job VM.
 BOSH communicates with the Agent, which in turn executes commands in the
 control script.
 The Agent does this using open source process monitoring software called
-[Monit](http://mmonit.com/monit/).
+[Monit, version 5.2.5](https://web.archive.org/web/20110816041503/https://mmonit.co+m/monit/documentation/monit.html).
```
selzoc, line 212: Looks like an extraneous `+` character made it in here (`mmonit.com+m`)

silvestre: Indeed, thanks for the eagle-eyed review!

ramonskie (approved):

rkoster (approved):

rkoster:
Thanks! @silvestre

PR #806, opened 2023-10-26
Update configs.md

fmoehler:
The documentation does not seem to be up to date for the Generic Configs, at least my cli behaves differently e.g. need to provide id as an argument when using delete-config command without specifying --type/--name, otherwise I get an error like 
`No configs to delete: no matches for id 'deploy' found.`


ystros (approved):

mvach (approved):

rkoster:
Thanks! @fmoehler

PR #815, opened 2023-11-16
Adding disaster recovery documentation for bosh in case of AZ outages

anshrupani:
We worked on making BOSH available also in case of AZ outages. With this doc, we want to share our experience and journey with the community regarding how we were able to achieve it.

### DIFF
--- a/content/disaster-recovery.md
+++ b/content/disaster-recovery.md
@@ -1,0 +1,99 @@
+# BOSH Disasters Recovery Process for AZ Outages
+
+[>Video-Presentation on the CF Day Heidelberg<](https://www.youtube.com/watch?v=0oMrGu9XuBY&list=PLhuMOCWn4P9jUHBucZBkSjmkwEbvx8vxf&index=12)
+
+Over the past months we invested a lot into a disaster recovery process, for our BOSH instances, that can be followed when
+a whole AWS Availability Zone is in an outage situation. As the BOSH director itself does not support a multi AZ setup
+it's a bit tricky to plug such a process from the outside into the director.
+
+Since BOSH is not relevant for the runtime, we don't expect that this process is to be executed if a zone is down for few minutes/hours.
+However after a while we most probably need to be able to deploy fixes for the existing BOSH deployments,
+and it would make sense to share the ideas with the community.
+
+In general, we deploy multiple BOSH directors per cloud. For simplicity, we assume two directors here. One of these
+directors (we name it `bootstrap-bosh`) is deployed via the [create_env](https://bosh.io/docs/cli-v2/#create-env) command.
+Its purpose is to deploy the second director (we name it `bosh`) instance, which is then used to deploy CF and other
+services. All our directors are deployed to `zone_1` by default.
+
+Our simple attempt in case of a zone outage is to deploy substitute `bootstrap-bosh` and `bosh` directors into one of the
+still available zones (let's assume `zone_2`).
+
+## Using domain names
+
+The very fist step in our journey to deploy a bosh director in another AZ was to use domain names instead of static IPs.
+Domain names would allow clients to use a static URL which is independent of the AZ. However, there were a few interesting scenarios
+we stumbled upon. These scenarios included updating certificates with DNS names along with thinking about the actual switch with
+which we would stop using static IPs and use URLs instead.
+
+## State externalisation
+
+Very early we decided to externalize as much state as possible before even thinking about a disaster recovery process.
+This required us to run the director on top of an external database (for example AWS RDS) and an external blobstore (for example AWS S3).
+
+- AWS RDS is replacing the Postgres database which is co-located on the director host by default.
+- AWS S3 is replacing the Nginx based blobstore which is as well co-located on the director host.
+
+Additionally, we decided that we don't care about the task logs that are stored on the persistent disk of the director
+host, instead we accept to lose such logs in case of an Availability Zone Outage.
+
+As we already use AWS RDS instances for all our relevant AWS Cloud installations, we just needed to start using an external blobstore.
+There are two options normally, either we accept to recreate our blobstore while switching to the external blobstore and bear with the longer blobstore
+recreation times and therefore longer duration for various bosh deployments for our production environments, or first migrate the blobs
+we have co-located on the director to S3 and then switch to using the external blobstore without recreating it.
+
+The first way is easy to implement, but costly with respect to the time it takes. It can easily be done by deploying 
+all bosh deployments with a "--fix-releases" flag. Though there's a risk of losing the DNS records stored
+locally as a blob within the director.
+
+We chose to implement the required blob migration from local Nginx blobstore to S3. In combination with the usage of
+signed URLs, doing the migration was tricky since one has to add the AWS Root CAs to the agent's blobstore
+CAs so that the agent can communicate during the migration with the local blobstore and S3 both at once.
+
+## Director isolation
+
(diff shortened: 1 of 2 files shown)


beyhan (changes requested):
I have to major points:

- we should focus on the technical part how this could be implemented and not so much on the setup SAP has.
- I think it is fine to stay with the `create-env` use case and describe what is required for it. 
- The Talk link we can move to the bottom for more information in case people want to follow it.

beyhan:
There was an update and this is again ready for review.

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+It would also make sense to use an external blobstore like AWS S3 which is highly available than consuming blobs from the colocated DAV blobstore. This would help to fail over to a new director with less worries of re-fetching the blobs (keeping in mind this is a time consuming process) when the need arises. As of now, it is already possible to use highly available blobstores provided by AWS, Azure, AliCloud and GCP. Switching over to a new blobstore can be smooth if there's a blobstore migration mechanism implemented using the existing BOSH Blobstore CLIs (by simply 'putting' the blobs using the s3cli from the director VM to the S3 bucket for example in case of AWS). Another way of doing this would be to recreate the blobstore externally by uploading the blobs with [--fix](https://bosh.io/docs/cli-v2/#upload-release) or running the deploy commands with [--fix-releases](https://bosh.io/docs/cli-v2/#deploy). This however is costly in terms of time required to re-fetch the blobs.
+See [this](https://bosh.io/docs/director-configure-blobstore) to start configuring the director to use an external blobstore.
+
+## Steps for recovering a bosh director in the event of an AZ outage:
+
```
beyhan, line 19: Could we describe the setup you have in mind here? I mean to say something like we habe deployments in `zone_1` and `zone_2`. The director is currently deployed in `zone_1`... It will help readers to understand the situation better.

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+### 1. Isolating the director
+To isolate the director in `zone_1` from the database to prevent it from modifying the state when the zone is up, it is important that:\
+a. The database passwords via terraform are rotated\
```
beyhan, line 22: I don't think that everyone uses terraform to setup the DB. Maybe we could be more generic here.

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+b. All existing database connections are killed via psql
```
beyhan, line 23: You could add a psql command here because this is quite specific. From where this should be executed? From the director VM itself?

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+See [this](https://bosh.io/docs/director-configure-blobstore) to start configuring the director to use an external blobstore.
+
+## Steps for recovering a bosh director in the event of an AZ outage:
+
+### 1. Isolating the director
+To isolate the director in `zone_1` from the database to prevent it from modifying the state when the zone is up, it is important that:\
```
beyhan, line 21: We should have an introduction here like. Let's assume that there is an outage in `zone_1`...

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+a. The database passwords via terraform are rotated\
+b. All existing database connections are killed via psql
+### 2. Deploying another director
+There might be differences in deploying BOSH directors in another AZ based on whether we use the `create-env` approach or the `bosh deploy` approach.
+
+After isolating the director deployed by `create-env` in `zone_1` the process to deploy a second one in `zone_2` can be treated similar to a scratch installation. That means that the zone used in the director's deployment manifest to is changed to `zone_2` and the `bosh-state.json` is dropped before calling [create-env](https://bosh.io/docs/cli-v2/#create-env) and setting up the second director in `zone_2`.
```
beyhan, line 27: We could simplify this sentence e.g. by: `That means the zone used in the director's deployment manifest is changed to `zone_2`...

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+
+For directors deployed by `bosh deploy`, the deployment manifest should be updated the used zone in the manifest should be set to `zone_2`. Next,the number of [instances](https://bosh.io/docs/manifest-v2/#instance-groups) in the `instance_group` has to be increased to 2 and a second static ip (available in `zone_2`) should be configured.
```
beyhan, line 29: `the used` most probably should be `to use`

beyhan, line 29: Why the instances should be increased? Which instances? What about the static ip?

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+
+However, while redeploying thi director with the updated manifest, the director which this drector is the deployment of would
+first try to delete the old `bosh` in `zone_1`. This will most probably fail in case of a zone outage. This can be prevented by invoking [ignore](https://bosh.io/docs/cli-v2/#ignore) on the deployed `bosh`
+in `zone_1`.
+### 3. Disabling resurrection
+After `bosh` is successfully deployed in `zone_2` it will now take care for repairing its deployments. Most of them would
+be in error situation since they would have instances in the failing zone. The director would continuously trigger `scan_and_fix`
+tasks which would then also fail. 
+Therefore, an option would be to deploy the new directors with with disabled resurrection with other monitoring techniques in place.
```
beyhan, line 38: Typo `with with`

beyhan, line 38: It is not completly clear what is meant by `...with other monitoring techniques in place.` We could leave this out or be more precise here.

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+After the zone outage is over, the directors in `zone_1` can be dropped and the ones in `zone_2` can be used further.
```
beyhan, line 39: Type `the ones` most probably should be `the one`

content/disaster-recovery.md
```diff
@@ -0,0 +1,43 @@
...
+Of course with enabled resurrection.
```
beyhan, line 40: We could add the command or links how resurrection could be enabled/disabled.

beyhan (changes requested):

content/disaster-recovery.md
anshrupani: I guess that in the case of a director (let's say B) that is deployed by another director A, A would have the original IP and instance records in it's database, and simply overwriting these records might lead to inconsistency or errors. Therefore we thought that it's probably better to configure two instances (and the other one also having another static IP).

anshrupani:
Thanks for your detailed comments @beyhan !
I have included all changes suggested by you.

content/disaster-recovery.md
```diff
@@ -0,0 +1,45 @@
...
+See [this](https://bosh.io/docs/director-configure-blobstore) to start configuring the director to use an external blobstore.
+
+## Steps for recovering a bosh director in the event of an AZ outage:
+
+We can assume that we are working with multiple deployments across two availability zones: `zone 1` and `zone 2`, and that our director is currently deployed in `zone 1`. Then, let's also assume that there's a sudden outage in  `zone 1`.
```
beyhan, line 20: Please make this consistent. Here we talk about `zone 1` and below `zone_1`. The same for zone 2.

beyhan (changes requested):

content/disaster-recovery.md
anshrupani: Oh, yeah sure!

jpalermo:
Thanks!

PR #826, opened 2024-02-16
Document credhub variables are not cleaned upon depl deletion

gberche-orange:
See related slack exchange at https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1706115913946389?thread_ts=1706106062.329369&cid=C02HPPYQ2

> I could not find any reference to delete in https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/config_server/client.rb (edited)


rkoster:
Thanks! @gberche-orange

PR #827, opened 2024-03-08
Resize Disk support docs for AWS & GCP

jdesulme:
It seems like the docs were never updated when these properties were enabled for both AWS & GCP. https://github.com/cloudfoundry/bosh-agent/pull/244


rkoster (approved):

rkoster:
Thanks! @jdesulme

PR #831, opened 2024-04-30
post-deploy scripts are _always_ run

cunnie:
post-deploy scripts for five years (since 268.7.0) were always run unless someone was motivated and clever enough to modify the generated `director.yml` job configuration.

Now post-deploy scripts are always run, and you can't change that behavior.


klakin-pivotal (approved):
L G T M

PR #842, opened 2024-06-19
Illustrate use of bosh curl in director API reference

gberche-orange:
Extra examples with use of the `bosh curl` command makes the API documentation actionable.

Currently, only added examples for GET methods. POST/DELETE examples may come in a later distinct PR.


lnguyen (approved):
LGTM

PR #843, opened 2024-06-19
Adding the CLI V2 reference also onto the Reference Tab

gberche-orange:
Was previously only on the guides Tab, despites legacy CLI v1 indeed being on the Reference tab (under Legacy Documentation)


fearoffish (approved):
This looks good to me, a simple documentation improvement. Thanks!

aramprice (approved):

PR #848, opened 2024-08-09
fix formatting for configure blobstore for gcp section

fmoehler:
There is a weird section in the docs:

https://bosh.io/docs/director-configure-blobstore/#properties-blobstore-provider-gcs-credentials_source-static-json_key-director-blobstore-service-account-file-bucket_name-test-bosh-bucket-storage_class-regional-agent-blobstore-credentials_source-static-json_key-agent-service-account-blobstore-file

also the yaml snippet appears bigger than for the other ones. Maybe this space helps?


beyhan (approved):

beyhan:
I checked with locally rendered docs that this helps.

PR #850, opened 2024-09-10
Native Disk Update

s4heid:
* marks the Native Disk Resize feature as supported on Azure (which was seemingly forgotten).
* marks the Native Disk Update feature as supported for Azure in the next CPI release (`v50.0.0`?)
* adds documentation which explains the newly introduced `update_disk` method
* moves the `disk_size` property outside of the `cloud_properties` block, as `disk_size` is not a cloud property.

### DIFF
--- a/content/aws.md
+++ b/content/aws.md
@@ -75,6 +75,7 @@ using the new key and migrate data onto the new disk.
 | --------------------------------- | ------------------------------ |
 | Multi-CPI                         | Supported, [v61][aws_cpi_v61]+ |
 | Native Disk Resize                | Supported, [v89][aws_cpi_v89]+ |
+| Native Disk Update                | Not Supported                  |
 | Generic VM Resource Configuration | Supported, [v56][aws_cpi_v56]+ |
 
 [aws_cpi_v61]: https://github.com/cloudfoundry/bosh-aws-cpi-release/releases/tag/v61
--- a/content/azure-cpi.md
+++ b/content/azure-cpi.md
@@ -326,20 +326,27 @@ instance_groups:
 ---
 ## Disk Types {: #disk-pools }
 
+* **name** [String, required]: Name of the disk type.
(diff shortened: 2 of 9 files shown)


MSSedusch (approved):
lgtm

jpalermo:
Should wait for https://github.com/cloudfoundry/bosh/pull/2555 before merging

aramprice (approved):

ystros (approved):

beyhan:
@rkoster, @aramprice I'm wondering what the status with the implementation of this is. It was reverted by https://github.com/cloudfoundry/bosh/pull/2570 because it failed the integration tests. Was there a revert of the revert?

aramprice:
@beyhan - the rollback of https://github.com/cloudfoundry/bosh/pull/2555 has not been undone.

As noted in https://github.com/cloudfoundry/bosh/pull/2570 we rolled back because the integration test that was added did not run (and in fact could not have run successfully given the way the test was written). To my knowledge our team hasn't done any work to fix the integration spec or re-merge https://github.com/cloudfoundry/bosh/pull/2555.

Unfortunately there isn't currently a documented way for folks to run the integration specs locally which makes it hard for contributors to write these tests. For folks who have write access to https://bosh.ci.cloudfoundry.org/ there is `rake fly:integration` which will execute the integration specs via Concourse, so depending on who is making the contribution this may be an option. Check out `src/bosh-dev/lib/bosh/dev/tasks/fly.rake` in the BOSH repo for more context.

beyhan:
@a-hassanin could someone of you who is an approval and has write access to https://bosh.ci.cloudfoundry.org/ support @s4heid  for running and debugging the integration test as suggested by @aramprice.

Another question I wanted to raise here is whether we should wait with the documentation until we have successfully integrated this feature. I think If we don't get the feature in the next week in we should revert the docs until we have it. @rkoster, @jpalermo any opinion on this?

PR #858, opened 2024-11-29
Adjust bosh pcap command documentation for bosh-cli#672

a18e:
Add example for multiple Instance Group parameters, `parallel` and `non-interactive` flags

Should be merged together with https://github.com/cloudfoundry/bosh-cli/pull/672


ramonskie (approved):
lgtm

PR #866, opened 2025-02-20
Fix styling for AZ migration warning

ystros:
Spotted a minor styling / readability problem on the AZs page due to bad markdown. Also specifically updated the note to be "danger" since data loss can occur if you attempt to move something with persistent disks across AZs.

Before:
Screenshot 2025-02-20 at 3 11 08 PM

After:
Screenshot 2025-02-20 at 3 09 17 PM

### DIFF
--- a/content/azs.md
+++ b/content/azs.md
@@ -139,7 +139,8 @@ VMs total: 4
     planning an AZ migration.
 
 
-!!! General limitations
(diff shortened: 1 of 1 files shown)


beyhan (approved):

beyhan:
Thank you @ystros! The change definitely adds the focus to the important points.

PR #903, opened 2026-05-12
Enhance update_disk documentation

Ivaylogi98:
## Summary

- Document that `update_disk` can return a new disk CID when the disk is replaced (e.g., snapshot-based type change on GCP). Previously documented as "No return value".
- Add section on Director behavior during VM recreation: the Director now calls `update_disk` while the disk is already detached, avoiding a redundant detach-update-attach cycle.
- Add API request/response examples showing both the disk-replaced and in-place-update cases.
- Add the Google CPI (`bosh-google-cpi-release`) to the Implementations list.

## Context

These changes reflect the new `update_disk` behavior introduced in:
- [cloudfoundry/bosh#2701](https://github.com/cloudfoundry/bosh/pull/2701) - Director persists new disk CID from `update_disk` return value and calls it during recreation
- [cloudfoundry/bosh-google-cpi-release#382](https://github.com/cloudfoundry/bosh-google-cpi-release/pull/382) - GCP CPI implementation of `update_disk` (snapshot-based type migration)

### DIFF
--- a/content/cpi-api-v2-method/update-disk.md
+++ b/content/cpi-api-v2-method/update-disk.md
@@ -14,6 +14,17 @@ Should the native disk update operation encounter limitations, indicated by a `B
 !!! note
     Before enabling this feature, ensure that the `director.enable_cpi_resize_disk` property is disabled, as the Director gives priority to the `resize_disk` method due to its broader implementation across CPIs. However, it's worth noting that the `update_disk` method not only matches the capabilities of `resize_disk` but also offers the additional functionality of updating cloud properties.
 
+## Director behavior during VM recreation
+
+When `enable_cpi_update_disk` is enabled and a VM is being recreated (e.g., due to a stemcell update or cloud property change), the Director calls `update_disk` while the disk is already in a detached state. This avoids a redundant detach-update-attach cycle:
+
+1. Detach disk from the old VM
+2. Call `update_disk` (disk is already detached)
+3. Delete old VM, create new VM
+4. Attach the (possibly new) disk to the new VM
+
+If the CPI does not support `update_disk`, the Director falls back to the standard copy-based disk update path transparently.
+
 ## Arguments
 
  * `disk_cid` [String]: Cloud ID of the disk to update; returned from `create_disk`.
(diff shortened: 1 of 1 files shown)


content/cpi-api-v2-method/update-disk.md
```diff
@@ -22,13 +33,36 @@ Should the native disk update operation encounter limitations, indicated by a `B
 
 ## Result
 
-No return value
+ * `new_disk_cid` [String or nil]: If the disk was replaced (e.g., via snapshot and recreate due to an incompatible type change), returns the new disk's Cloud ID. The Director persists this new CID and uses it for subsequent operations. Returns `nil` if the disk was updated in-place (e.g., a size-only change).
```
neddp:
```suggestion
 * `new_disk_cid` [String or nil]: If the disk was replaced (e.g., via snapshot and recreate due to an incompatible type change), returns the new disk's Cloud ID. The Director persists this new CID and uses it for subsequent operations. If the disk was updated in-place, the CPI may return `nil` or the original disk CID - in both cases, the Director continues using the existing CID.
```

content/cpi-api-v2-method/update-disk.md
```diff
@@ -14,6 +14,17 @@ Should the native disk update operation encounter limitations, indicated by a `B
...
+## Director behavior during VM recreation
+
+When `enable_cpi_update_disk` is enabled and a VM is being recreated (e.g., due to a stemcell update or cloud property change), the Director calls `update_disk` while the disk is already in a detached state. This avoids a redundant detach-update-attach cycle:
+
+1. Detach disk from the old VM
+2. Call `update_disk` (disk is already detached)
+3. Delete old VM, create new VM
+4. Attach the (possibly new) disk to the new VM
+
+If the CPI does not support `update_disk`, the Director falls back to the standard copy-based disk update path transparently.
```
neddp:
```suggestion
If the CPI does not support `update_disk`, the Director falls back to the standard copy-based disk migration path transparently.
```

aramprice (approved):

neddp (approved):

PR #913, opened 2026-07-10
ci: fix `build.yml` task mkdocs path

achrinza:
Fixes this issue caused by the `mkdocs-material` upgrade.

see: https://github.com/cloudfoundry/docs-bosh/pull/908
fixes: https://github.com/cloudfoundry/bosh-io-web/pull/81#issuecomment-4929539085


neddp (approved):
Thanks!

aramprice:
Thank you (all) for the quick turnaround!

